PR #9, opened 2015-07-16
Introducing ReplyMarkup.php entity

MBoretto:
I've wrote this class to implement easily interfaces with reply markup.
It support Keyboard, Hidekeyboard and ForceReply.
I've also attached some new function where is possible to see how use this class into the commands.
The code passed the test described in the comtributing.md


akalongman:
@MBoretto why closed?

MBoretto:
I've merger ReplyKeyboardMarkup ReplyKeyboardHide and Force Reply in one class. Maybe it's better to build three differet classes.
Then i build a custom function to generate json string instead of use the one inherited from Entities. I woul like to rewrite it. That's all!

akalongman:
Ok :) I appreciate your help

PR #20, opened 2015-08-08
Fix help new chat participant work

MBoretto:
I've fix the /help bug now works properly.
Finally the bot trigger the message new chat participant and left chat particiapant.


MBoretto:
You are right, let me fix it!

PR #24, opened 2015-08-24
[Fix] check if is a command before tock

MBoretto:
I've fixed a little bug. Not all messages to the  bot are commands, some of them (messages from the keyboard) can be pure text only. 
I've change in the methods getFullCommand  and getText to support pure text messagges.


akalongman:
Closed? Why?

PR #55, opened 2015-11-29
Do not print debug data

boukeversteegh:


MBoretto:
Thanks for contributing!

PR #56, opened 2015-11-29
Fix fatal error in NewchatparticipantCommand.php

TheReverend403:


akalongman:
@TheReverend403 Thanks for contributing!

PR #76, opened 2016-01-25
Initial code cleanup of `Telegram.php`.

noplanman:
No functional changes, only slight modifications of documentation and code layout.
This commit is mainly to discuss with the project members if this style is ok and can be further refined.

@MBoretto, the discussion is opened :smiley:

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -23,7 +23,6 @@ use Longman\TelegramBot\Exception\TelegramException;
  */
 class Telegram
 {
-
     /**
      * Version
      *
@@ -67,7 +66,7 @@ class Telegram
     protected $update;
 
     /**
-     * Log Requests
+     * Log requests
      *
      * @var bool
      */
@@ -81,14 +80,14 @@ class Telegram
     protected $log_path;
 
     /**
-     * Upload Path
+     * Upload path
      *
      * @var string
      */
     protected $upload_path;
 
     /**
-     * Dowload Path
+     * Download path
      *
      * @var string
      */
@@ -102,7 +101,7 @@ class Telegram
     protected $log_verbosity;
 
     /**
-     * MySQL Integration
+     * MySQL integration
      *
      * @var boolean
      */
@@ -127,14 +126,15 @@ class Telegram
      *
      * @var array
      */
-
-    protected $message_types =['Message', 'Photo', 'Audio', 'Document', 'Sticker', 'Video',
+    protected $message_types = [
+        'Message', 'Photo', 'Audio', 'Document', 'Sticker', 'Video',
         'Voice', 'Location', 'command', 'new_chat_participant',
         'left_chat_participant', 'new_chat_title', 'delete_chat_photo',
         'group_chat_created', 'supergroup_chat_created', 'channel_chat_created',
-        ];
+    ];
+
     /**
-     * Admins List
+     * Admins list
      *
      * @var array
      */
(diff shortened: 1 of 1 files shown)


src/Telegram.php
```diff
@@ -23,7 +23,6 @@
  */
 class Telegram
 {
-
```
noplanman: Some classes have an empty line here, others don't. To me no empty line looks better.
If there should be an empty line, should there also be one at the very end of the class?

src/Telegram.php
```diff
@@ -127,14 +126,15 @@ class Telegram
      *
      * @var array
      */
-
-    protected $message_types =['Message', 'Photo', 'Audio', 'Document', 'Sticker', 'Video',
+    protected $message_types = [
+        'Message', 'Photo', 'Audio', 'Document', 'Sticker', 'Video',
         'Voice', 'Location', 'command', 'new_chat_participant',
         'left_chat_participant', 'new_chat_title', 'delete_chat_photo',
         'group_chat_created', 'supergroup_chat_created', 'channel_chat_created',
```
noplanman: This array's last item ends with a comma. Other arrays don't.
Is this intentional or a forgotten comma?
Personally I prefer having a leading comma, as it prevents errors when reordering array code for example.

src/Telegram.php
```diff
@@ -396,32 +389,31 @@ public function handle()
...
                 }
             }
-    
+
             // check type
             $message = $update->getMessage();
             $type = $message->getType();
-    
+
             switch ($type) {
```
noplanman: All of these command types, apart from the "Generic" ones, follow the same conversion rules.
- Remove "_"
- Capitalise first character
For that reason, this part can be shortened by quite a bit.

At the same time I can see that having each command on a separate `case`, additional code could be added on a case by case basis. Although this could also be done for individual cases specifically: `if..then..`

Shall I attempt to shorten this function or do you want to leave it as it?

src/Telegram.php
MBoretto: We can remove empty line from the begin and the end of the class.

src/Telegram.php
MBoretto: Is not intentional, we can leave coma and benefit if reordering as you said.

src/Telegram.php
MBoretto: It would be nice shrink!
But let me explain some issue to overcome!

There are two kind of commands:
- Command about chat trigger as new_chat_participant...
- Commands about type of message as Video, Voice...

Each one belong to first type has is own command, instead the latter type is completely redirected to Generic.
I wonder if can be a  good idea to separate those two kind of commands in this way:

##Current configuration
Message entity has [type](https://github.com/akalongman/php-telegram-bot/blob/master/src/Entities/Message.php#L75) member that can be: new_chat_participant, ... , command, Video, .. , Message ecc.

##New configuration
Message entity with two new member:
- *trigger*: new_chat_participant, Message, Command, Left_chat_participant, (all the chat trigger + command and message )
- *message_type*: Video, Photo, text, Sticker, null if *trigger* is  not Message or Command

In this way we can have a command for each *trigger* and we can handle different type of messages inside the GenericmessageCommand.
What do you think? 

Ah! Of course we can directly return the command result! :)

src/Telegram.php
noplanman: Ok :+1:

src/Telegram.php
noplanman: Ok :+1:

PR #78, opened 2016-01-31
Initial code cleanup of `DB.php`.

noplanman:
No functional changes, only slight modifications of documentation and code layout.

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -38,12 +38,12 @@ class DB
     /**
      * PDO object
      *
-     * @var \PDO
+     * @var PDO
      */
     static protected $pdo;
 
     /**
-     * table prefix
+     * Table prefix
      *
      * @var string
      */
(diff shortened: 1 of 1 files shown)


MBoretto:
-Ok for alphabetical ordering!
-Ok for No empty line after <?php
-OK for 1 empty line after the file header

We can start inserting a simple header for each class like:
```
/**
 * Class ClassName.
 */
```
And leave the default header for each file! @akalongman  Do you agree with that?

I have a question, the file header is formatted  with:
```
/*
 *
 *
 */
```
Should instread like the other dockblock?
```
/**
 *
 *
 */
```

akalongman:
No empty line after <?php ?
Please, follow the PSR-2 coding standard

akalongman:
If php-cs passes with no notices, everything is ok

MBoretto:
It seems so from the travis checks

akalongman:
Nice. You are doing amazing job guys

noplanman:
@akalongman Does the PSR-2 require an empty space after the `<?php`?
I've searched for PSR-2 "guides" but couldn't find anything regarding that.

Could you recommend a thorough read for me to learn? :blush:

Thanks!

noplanman:
@MBoretto Yes, DocBlock format (`/**`) should be used for the headers, otherwise phpdoc will give a nice red bar saying `No summary was found for this file`.

Ok for leaving default file header!

PR #81, opened 2016-02-03
Code cleanup for TelegramException.php

noplanman:


src/Exception/TelegramException.php
```diff
@@ -1,17 +1,26 @@
...
 class TelegramException extends \Exception
 {
-
+    /**
+     * Exception constructor that writes the exception message to the logfile
+     *
+     * @param string  $message Error message
+     * @param integer $code    Error code
+     */
     public function __construct($message, $code = 0)
```
MBoretto: It would be fine to add a method to enable and disable exception log, maybe with a defining a var or a global variable?

noplanman: We could add a `static` variable for example.
And maybe add another one to choose the path/filename where the file gets written?

MBoretto: Yes it sounds good, then we should initialize in Telegram.php

PR #83, opened 2016-02-03
Code cleanup for `Commands`

noplanman:
This PR is just for filling up the documentation and modifying some code style.
The commands in general need a bigger overhaul to prevent duplicate code and make it more sensible.

### DIFF
--- a/src/Commands/GenericCommand.php
+++ b/src/Commands/GenericCommand.php
@@ -64,15 +64,17 @@ class GenericCommand extends Command
         $update = $this->getUpdate();
         $message = $this->getMessage();
         $chat_id = $message->getChat()->getId();
-        //you can use $command as param
+        //You can use $command as param
         $command = $message->getCommand();
  
         $chat_id = $message->getChat()->getId();
         $text = $message->getText(true);
 
-        $data = array();
-        $data['chat_id'] = $chat_id;
-        $data['text'] = 'Command: '.$command.' not found.. :(';
+        $data = [
+            'chat_id' => $chat_id,
+            'text'    => 'Command: ' . $command . ' not found.. :(',
+        ];
+
         $result = Request::sendMessage($data);
         return $result->isOk();
     }
--- a/src/Commands/GenericCommand.php
+++ b/src/Commands/GenericCommand.php
@@ -1,27 +1,61 @@
 <?php
-
-/*
+/**
  * This file is part of the TelegramBot package.
  *
  * (c) Avtandil Kikabidze aka LONGMAN <akalongman@gmail.com>
  *
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
-*/
+ */
+
 namespace Longman\TelegramBot\Commands;
 
-use Longman\TelegramBot\Request;
 use Longman\TelegramBot\Command;
 use Longman\TelegramBot\Entities\Update;
+use Longman\TelegramBot\Request;
 
 class GenericCommand extends Command
 {
+    /**
+     * Name
+     *
+     * @var string
+     */
     protected $name = 'Generic';
-    protected $description = 'Handle generic commands or is executed by default when a command is not found';
+
+    /**
+     * Description
+     *
+     * @var string
+     */
+    protected $description = 'Handles generic commands or is executed by default when a command is not found';
+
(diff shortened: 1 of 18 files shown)


src/Commands/SlapCommand.php
```diff
@@ -28,20 +66,22 @@ public function execute()
         $chat_id = $message->getChat()->getId();
         $message_id = $message->getMessageId();
         $text = $message->getText(true);
-        
-        $sender='@'.$message->getFrom()->getUsername();
 
-        $data = array();
-        $data['chat_id'] = $chat_id;
+        $sender = '@' . $message->getFrom()->getUsername();
```
MBoretto: We should use the methods ->tryMention() from the class User.
https://github.com/akalongman/php-telegram-bot/blob/master/src/Entities/User.php#L63

src/Commands/SupergroupchatcreatedCommand.php
```diff
@@ -1,27 +1,66 @@
...
+     * If this command is enabled
+     *
+     * @var boolean
+     */
     protected $enabled = true;
 
+    /**
+     * Execute command
+     *
+     * @todo $chat_id isn't defined!
```
MBoretto: You are right this is an error..

src/Commands/GenericCommand.php
```diff
@@ -1,41 +1,83 @@
...
+     * If this command is enabled
+     *
+     * @var boolean
+     */
     protected $enabled = true;
 
+    /**
+     * Execute command
+     *
+     * @todo This can't be right, as it always returns "Command: xyz not found.. :("
```
MBoretto: This command is called with recursion here: 
https://github.com/akalongman/php-telegram-bot/blob/master/src/Telegram.php#L492
After that all the commands in the dirs has been parsed.

MBoretto: Did you noticed that you can set if a command is public or not:
https://github.com/akalongman/php-telegram-bot/blob/master/src/Command.php#L24
This option is used to show or hide commads in  the /help 
https://github.com/akalongman/php-telegram-bot/blob/master/src/Commands/HelpCommand.php#L45

I think that this should be extended also to the execution. So public command can be executed by the user instead private one (As generic, Newchatparticipant, ... and so on) should be executed only by the library.

noplanman: In my mind (as far as I've understood the different commands so far), there should be 3 different kinds of commands:
- Executed by the library / internal system
- Executed by the Admins
- Executed by the Users (& Admins obviously)

This way, we could separate these different commands more easily, as they have their own specific way of being executed. I see at the moment that many commands do pretty much the same thing, but each one uses lots of identical code, which really should be optimised.

Yes I saw that @ public vs. private. I think public should be the ones all Users can execute and private the ones only Admins can.

I haven't looked deeper into this, but maybe adding an extra layer could help out a lot:
```
-> Command
  |-> AdminCommand
    |-> SendToAllAdminCommand
  |-> UserCommand
    |-> WhoAmIUserCommand
  |-> SystemCommand
    |-> NewChatTitleSystemCommand
```

Something like that, you know what I mean?

MBoretto:
Yes I think that this separation is needed.
With this schema we don't need anymore public variable.
System commands are not shown in the /help.
User commands  are visible to user in /help
Admin can see both user and admin commands in /help

noplanman:
Yes. Also, I feel it's easier for developers to understand the different types of command.

Am I right that the System commands are given by Telegram?
Developers can't write their own, right?

MBoretto:
They are connected to some chat action like `new_chat_participant` so they are given by telegram.
Anyway we should provide the possibility to overwrite the default one with their own. 
(this can be achieved with the custom folder commands).
For example genericcommand can be used to cought commands as param like /1 /green /the option chosen

noplanman:
> possibility to overwrite the default one with their own

You mean when a System command like `new_chat_participant` is called, a developer can execute his own custom code? If so, absolutely!

Regarding the `GenericCommand`, that would be a `UserCommand`, right? So that any user can call `/whateveriwant` or `/maybeeventhis` and the developer can decide what happens to them. This would of course only be the case, if no separate command explicitly handles them.

Have I understood correctly?

PR #97, opened 2016-02-10
Just inherit repeating documentation blocks

noplanman:
Instead of having to write the same docblocks over and over again for each command, just inherit them for the parent class!

Unfortunately, when generating the documentation, phpdoc gives a notice about not finding the documentation for the member variables, because at compilation time it's impossible to know if it can inherit the values.
That put aside, it's just SO much easier to read the code without tons of comments everywhere.


noplanman:
@MBoretto Just cleaned up the conflicts, sorry about that. Will try to split up the changes better next time.

PR #98, opened 2016-02-10
Remove any unnecessary code from commands.

### LINKED ISSUE #86: /SupergroupchatcreatedCommand
chat_id is not defined should be fixed.

noplanman:
Changes include:
- Removed the definitions of member variables that had the same values as their inherited defaults. These can be removed since the default values apply anyway.
- Removed superfluous code form execute command.

Fixed #86 in the process


MBoretto:
Your job is amazing! well done!

PR #101, opened 2016-02-16
Excody fix

MBoretto:
Fetched fix from @excody .
Now newchat participant and left one are stored propely


noplanman:
@Excody, great find, thanks a lot! :+1: 

@MBoretto Looks good, shall I call the merge train to let this PR aboard? :wink:

PR #102, opened 2016-02-16
Introducing db external connection

MBoretto:
Let you use DB.php with an external pdo connection


vitalyzhakov:
Hello!
How I can use Postgres?

MBoretto:
Seems supported from pdo !

chuv1:
SQLite supported by PDO too, but I had no luck to make it work...
So MySQL is the only option, right?

jacklul:
Wouldn't using external pdo connection be a 'hacky' way to use SQLite?

Issue #87, opened 2016-02-04
New command structure


MBoretto:
As @noplanman suggest in a comment commands structure can move to this configuration:

Copied form @noplanman commet:
In my mind (as far as I've understood the different commands so far), there should be 3 different kinds of commands:

Executed by the library / internal system
Executed by the Admins
Executed by the Users (& Admins obviously)
This way, we could separate these different commands more easily, as they have their own specific way of being executed. I see at the moment that many commands do pretty much the same thing, but each one uses lots of identical code, which really should be optimised.

Yes I saw that @ public vs. private. I think public should be the ones all Users can execute and private the ones only Admins can.

I haven't looked deeper into this, but maybe adding an extra layer could help out a lot:
```
-> Command
  |-> AdminCommand
    |-> SendToAllAdminCommand
  |-> UserCommand
    |-> WhoAmIUserCommand
  |-> SystemCommand
    |-> NewChatTitleSystemCommand
```

(my opinion)
In this way we can remove the public var ad show them according to the division
- **System commands** are not shown in the /help.
- **User commands** are visible to user in /help
- Admin can see both **user and admin commands** in /help

Lets start the discussion!

noplanman:
What I've been thinking about in terms of Commands, is how a user can totally define his/her own ones, removing the existing ones the library brings along.
The issue is, that the `Commands` directory is always included and can't be ignored.

Let me illustrate with an example:
I want to create a very specific bot that only supports 3 commands, say `/yesterday`, `/today` and `/tomorrow`.

Ok, I create a new directory with those commands in them and add it as a commands directory, fine.

Also, I want to provide an `InlineQuery` for a special `/now` command, which returns a list of items selected from the DB.

How would I do this with the current library without touching any library code?

MBoretto:
I think that we should implement a method that let you enable/disable certain commands during the setup. Then when you include a new dir of commands you can overwrite a present one if they have the same name

noplanman:
It would be really cool to be able to implement something like the [hooks](http://codex.wordpress.org/Plugin_API#Hooks.2C_Actions_and_Filters) that WordPress uses.
Hooks allow a plugin or theme developer to "hook into" core functionality very very easily.

Given the example here, there would be a hook to set the command directories. This way a user could easily "hook into" the Telegram class and set / unset any custom command directories. The same method could be applied to enable / disable a list of commands.

About simply overriding existing commands, that's how it works at the moment already! :smiley: 
[...]

PR #103, opened 2016-02-23
New command structure alpha

noplanman:
Initial PR to continue discussion started in #87 

**This PR needs a thorough code review first before being merged!**

### Summary of main changes
- Commands have been split up into 3 groups: **System**, **Admin** and **User**.
- Custom commands included by the user override the default ones. If multiple folders with custom commands get included, the latest include takes preference.
- Admin commands with the same name as User commands take preference when called by an admin.
- Commands now return a `ServerResponse` object to offer more control after the command has executed (also used for testing commands).

(For a more detailed list of changes, follow the thread in #87, especially [this comment](https://github.com/akalongman/php-telegram-bot/issues/87#issuecomment-186854364))


src/Commands/UserCommands/WhoamiCommand.php
```diff
@@ -10,16 +10,16 @@
  * Written by Marco Boretto <marco.bore@gmail.com>
  */
```
MBoretto: The command return ->isOk() insetad of serverResponse

noplanman:
Right you are, I missed that one, thanks! :blush:

MBoretto:
I also checked custom commands extension and everything works!  well done man!

PR #105, opened 2016-02-26
Empty response

MBoretto:
Added the method emptyResponse to use in command that don't need to fire a Request to telegram.


noplanman:
Looks good, merging.

Issue #106, opened 2016-02-26
Refactor Telegram.php


noplanman:
This issue is intended as a discussion board and checklist of tasks to refactor the `Telegram` class.
(*new class names are just suggestions and prone to change*)

The main objective is to separate the responsibility and to decouple as much as possible.
The `Telegram` class should handle only the main Telegram API related functionality and thread all sub-processes together by saving necessary objects in member variables, making them available to modify certain features. (i.e. having a `$commander` variable that handles all "command" functionality, etc.)

- [ ] Extract all **command** related functionality to a dedicated `Commander` class.
- [x] Extract all **logging** related functionality to a dedicated `Logger` class.

MBoretto:
It would be nice to separate also the log part.
Here's some thougts about:
- Logging should handle Request log as incoming update and also exception log in order to fix https://github.com/akalongman/php-telegram-bot/issues/96
- Since some months at every version the table schema change. Sorry for this.. No one complain about this but is a pain move existing data from a schema to another. In order to overcome this problem I was thinking that the incoming log update can be used to refill the database when the schema change. What you think about this?
- For debugging purpose was useful to me to have a single log file where i can see both incoming update and curl response of the library (according to log level). This feature is not good if we want to implement database refilling. We should find a way to handle those things together (If we decide to develop both)

noplanman:
How exactly do you envision the database refilling?
I don't understand how it's connected to the logging :confused:
Please do explain more :smiley: 

Have you thought of having a dedicated upgrade file that can be executed to update the database structure? Then, each time the schema is updated, the user can just run the "update" script and all the changes are made automatically. All we need to do then, is have a way to remember which database version we're on, so that only the new changes are applied.

MBoretto:
Incoming update (json string from webhook and getUpdates) can be logged in a text file with those options with the methods:
```
$telegram->setLogRequests(true);
$telegram->setLogPath($BOT_NAME . '.log');
```

So if you store `all` the incoming update in a textfile you can build a mechanims than read them all and reprocess it without send request back. In this way you can refill database.

noplanman:
Aaaaah, now I understand what you mean!

So basically what you're saying, is to save **all** updates to a separate file, which can be used as a data import on the database?

Is this necessary? Wouldn't it be enough to just track the database schema changes and update the database after each new version?
For an active bot, the import log file could become really big and the use can disable it, which brings us to the same problem :confused:

MBoretto:
[...]

Issue #96, opened 2016-02-10
The path used for writing to `TelegramException.log` is a security risk


dstotijn:
The `TelegramException` class writes a log file `TelegramException.log` in the current path of the executing PHP script, because `file_put_contents` is called with just a filename. This can be a security risk when a public facing web server is in place, where the log file could be in the document root and can be accessed.

noplanman:
You're absolutely right, thanks for pointing this out!

An easy fix would be to restrict access to the file using .htaccess or within the nginx configuration.

Other option would be to offer the option to choose the destination of the `TelegramException.log` file. This would require the user to set the destination and actually make sure that it can't be accessed.

#88 Might need to be implemented after all.

Also, we'll need to add a simple .htaccess / nginx config snippet to the readme, so that users can protect themselves.

ivanbaldo:
Maybe this could be closed now?
Commit https://github.com/akalongman/php-telegram-bot/commit/b1184d862d17fb8e300d57944658a9608742f904 deals with this.

MBoretto:
yes thanks!

PR #108, opened 2016-02-27
fix Exception Log

MBoretto:
I introduced the logger Class https://github.com/akalongman/php-telegram-bot/issues/106 to Handle the path of the exception logs.
Don't want to rewrite log system now just path this bug https://github.com/akalongman/php-telegram-bot/issues/96 .
Exception are not stored by default, you can start log then adding:
```
Longman\TelegramBot\Logger::initialize('Exceptionlog.log');
```
in your webhook.php or in your getUpdates.php

### DIFF
--- a/README.md
+++ b/README.md
@@ -427,8 +427,11 @@ $telegram->setUploadPath('yourpath/Upload');
 ```
 
 ### Logging
+Thrown Exceptions are not stored by default. You can Enable this feature adding this line in your 'webhook.php' or 'getUpdates.php'
 
-Thrown Exceptions are stored in *TelegramException.log* file (in the base directory).
+```php
+    Longman\TelegramBot\Logger::initialize('your_path/TelegramException.log');
+```
 
 Incoming update (json string from webhook and getUpdates) can be logged in a text file. Set those options with the methods:
 ```php
--- a/src/Exception/TelegramException.php
+++ b/src/Exception/TelegramException.php
@@ -10,6 +10,8 @@
 
 namespace Longman\TelegramBot\Exception;
 
+use Longman\TelegramBot\Logger;
+
 /**
  * Main exception class used for exception handling
  */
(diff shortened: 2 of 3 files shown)


noplanman:
Great!

About the logging, are we reducing the logs to just 1 file now?
Before we had a file for "Logs" and one for "Exceptions".

With this logger there will be only 1 file that combine both, yes?
Do we want to keep them separate or put all logs together?

The advantage of keeping them separate is that users can simply send us a copy of the exceptions without worrying about sharing private information like bot name or API key, etc.
So it would be great if there were still both. The Logger class manage them all of course.

What do you think?

MBoretto:
This class handles log for exception while the request one are still menaged by Request class. 
Log path are configurable, so if user want can put everything in one file.
I prefer to have separate path for different logs, this also to build the refill feature for the database https://github.com/akalongman/php-telegram-bot/issues/106#issuecomment-189233240

noplanman:
So you don't think it would make sense to put all logging functionality together into the Logger class?
My initial idea was to have a class that handles / manages all logging, normal logs and exceptions.
This is to decouple the logging and ease up the configuration.

MBoretto:
In the future only one class will handles all logs i totally agree! 
Maybe we can refactor this in the next release 0.29. Instead in 0.28 we can just be satisfy about this double logging system! What do you think?

noplanman:
To me it makes more sense to add a feature when it's thought through more, otherwise it may change again in the next version, which means the user will again need to update their code.
I'd suggest to leave the logging as it is now and maybe just refactor the logging for version 0.29.

It's better to add a "small" feature that works 100% in a separate release, than add bits of it in one and then need to change it again for the next.

MBoretto:
We could exploit [monolog](https://github.com/Seldaek/monolog) to do this. What do you  thing @akalongman @noplanman  ?

noplanman:
I think that for the logging that we'd be doing a library like that is overkill.
The usage does look nice and simple though ;-)

We're just writing to 2 files, right? One for the Telegram responses and cURL logs, and one for the Exceptions.

An advantage would be that user commands could also include logging very easily, which may be the case!
Also, users could define their own log destinations in the hook.php file, which would be pretty cool.

Hmm, I'm liking the idea of this library more and more as I'm typing this :smile:

akalongman:
I Hightly recommend use monolog.
@noplanman it has many advantages. We dont need test logging behaviour, everything is tested. We dont need implement wotking with filesystem, already exists and etc.

PR #111, opened 2016-02-28
Create MySQL database with correct table structure.

noplanman:
Having the database set up is necessary for certain tests which need to be implemented.


MBoretto:
good!

PR #119, opened 2016-03-29
Add new_chat_photo command

jacklul:
Noticed this hasn't been added yet, I know this can be handled by genericmessage but having comands for each message type might be a good idea.


MBoretto:
Yes I agree, good catch!

PR #120, opened 2016-03-29
Add migrate_to_chat_id and migrate_from_chat_id commands

jacklul:
Code in **SupergroupchatcreatedCommand.php** will never run as intended, **supergroup_chat_created** works just like **group_chat_created**, **migrate_to_chat_id** and **migrate_from_chat_id** are sent seperately.

Placing code from **SupergroupchatcreatedCommand.php** into **MigratetochatidCommand.php** and editing it a bit will also don't work, bot can't message new group id answering to update from old group id. (tried to figure it out and it just doesn't work)

Also **migrate_from_chat_id** is not always sent to bot.

### DIFF
--- a/src/Commands/SystemCommands/MigratefromchatidCommand.php
+++ b/src/Commands/SystemCommands/MigratefromchatidCommand.php
@@ -1,0 +1,36 @@
+<?php
+/**
+ * This file is part of the TelegramBot package.
+ *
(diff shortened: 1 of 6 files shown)


jacklul:
Also DB class:
'chat' table should be instantly updated with new entry when migrate_to_chat_id is received (e37eb58 is just my proposition)
Also some params were set in a weird way?

Chat id migration must be looked into, there might be more mistakes in code.

PR #130, opened 2016-04-11
Improved 'chats' command

jacklul:
List of changes done:
-all chats are no longer printed by default
-added search by group/user name and also user's first/last name
-chat id will be also shown
-fixed an issue where super groups were listed as normal groups

Showing all chats by default isn't good idea considering someone can have like ~1000 entries in DB, it will take some time and can prevent bot from answering to requests from other people, not to mention the message spam in the conversation.


MBoretto:
Thanks! @jacklul

PR #133, opened 2016-04-12
Add /whois command

jacklul:
This command will allows to look up basic info about user/group stored in DB.
Chats command will display shortcuts to it instead of id (only if whois is available).

### DIFF
--- a/src/Commands/AdminCommands/ChatsCommand.php
+++ b/src/Commands/AdminCommands/ChatsCommand.php
@@ -41,9 +41,9 @@ class ChatsCommand extends AdminCommand
         $text = trim($message->getText(true));
 
         $results = DB::selectChats(
-            true, //Send to groups (group chat)
-            true, //Send to supergroups (super group chat)
-            true, //Send to users (single chat)
+            true, //Select groups (group chat)
+            true, //Select supergroups (super group chat)
+            true, //Select users (single chat)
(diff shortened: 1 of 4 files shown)


jacklul:
Additionally optimized this and 'chats' a bit, it's better to just select one row instead of LOTS OF THEM and then iterating over them.

noplanman:
Cool, I really like this command! 👍 
Would be great if `/whois @noplanman` or equivalent would work too, not only via ID 😊

I also saw that your new `/chats` command returns the link without a space after the command (e.g. `/whois12345` instead of `/whois 12345`). Is this intentional? Both of those commands work for users though.
For groups it is `/whoisg12345` that works, whereas `/whois g12345` doesn't.

jacklul:
>I also saw that your new /chats command returns the link without a space after the command (e.g. /whois12345 instead of /whois 12345). Is this intentional?

Yes, so it can be clicked on /chats list.

>For groups it is /whoisg12345 that works, whereas /whois g12345 doesn't.

That g is just a replacement for '-' and is only there so the command can be clicked.
```/whois g12345``` is not supposed to work, use ```/whois -12345``` instead!

>Would be great if /whois @noplanman or equivalent would work too, not only via ID

Good idea!

jacklul:
Cleaner branch commit: https://github.com/jacklul/php-telegram-bot/commit/28bd0f91e86b8ab35525fafaea79d3e2bb98da71

PR #134, opened 2016-04-13
Throw on invalid responses

antonioribeiro:
If you send something inexistent to Telegram it will return an HTML 302 response. So now people should be able catch those exceptions:

image

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -317,8 +317,13 @@ class Request
             return new ServerResponse($fake_response, $bot_name);
         }
 
-        $response = self::executeCurl($action, $data);
-        return new ServerResponse(json_decode($response, true), $bot_name);
+        $response = json_decode(self::executeCurl($action, $data), true);
+
+        if (is_null($response)) {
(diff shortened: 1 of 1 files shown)


MBoretto:
Could you explain more in details the point? i'm not getting it

Then for what concern the travis fail you should perform all the test as described in https://github.com/akalongman/php-telegram-bot/blob/master/CONTRIBUTING.md

antonioribeiro:
If you set your bot name to 'xxxxxxx' and your bot api token to 'yyyyyy' you should get this response from `self::executeCurl($action, $data)`:

	<html>
		<head><title>302 Found</title></head>
		<body bgcolor="white">
			<center><h1>302 Found</h1></center>
			<hr><center>nginx/1.6.2</center>
		</body>
	</html>

This PR is only to throw the proper exception (TelegramException) in case someone else does something as dumb as that.

Here's a screenshot of the response being debbuged:

image

MBoretto:
Thank you for contributing! I will close here because I fetch your commit to merge it on develop! :;

antonioribeiro:
:+1:

PR #136, opened 2016-04-14
for UTF-8 characters in the class definition and commands

CrazyRedScorp:
Support for unicode characters in commands.


noplanman:
It looks like your editor uses Windows style line endings.
Please correct the line endings to Unix style, so that we can see what you've done here, thanks.

PR #139, opened 2016-04-14
Update ImageCommand.php

sndnvaps:
Return Random Picture from the UploadPath();


MBoretto:
Thanks for contributing could you push on develop?

sndnvaps:
ok . i will push to develop now .

PR #140, opened 2016-04-14
Update ImageCommand.php

sndnvaps:
Return Random Picture from the UploadPath();


MBoretto:
Thanks for contributing!

PR #144, opened 2016-04-14
Update Telegram.php

CrazyRedScorp:
Support for unicode characters in commands. _(try number three, edit file in github editors)_


CrazyRedScorp:
What a...

PR #147, opened 2016-04-15
Update /src/Commands/UserCommands/WeatherCommand.php

CrazyRedScorp:
Recently http://openweathermap.org/ added "API key" in user profile.
Without API key the system reports an error - "Can not find weather for location: <city>"

### DIFF
--- a/src/Commands/UserCommands/WeatherCommand.php
+++ b/src/Commands/UserCommands/WeatherCommand.php
@@ -37,7 +37,8 @@ class WeatherCommand extends UserCommand
      */
     private function getWeather($location)
     {
-        $url = 'http://api.openweathermap.org/data/2.5/weather?q=' . $location . '&units=metric';
+        $appid = 'Your_APIkey_from_openweathermap.org';
+        $url = 'http://api.openweathermap.org/data/2.5/weather?q=' . $location . '&units=metric&appid=' . $appid;
 
         $ch = curl_init();
         $curlConfig = [


src/Commands/UserCommands/WeatherCommand.php
```diff
@@ -37,7 +37,8 @@ class WeatherCommand extends UserCommand
      */
     private function getWeather($location)
     {
-        $url = 'http://api.openweathermap.org/data/2.5/weather?q=' . $location . '&units=metric';
+        $appid = 'Your_APIkey_from_openweathermap.org';
```
MBoretto: You can't ash  user  to edit file inside the  library you should set it with: https://github.com/akalongman/php-telegram-bot#commands-configuration

MBoretto:
We can't ask user to edit the file with the token, maybe we can find another api for forecast that doesn't require the  authentication. Is really not working anymore? Because sometimes the service seem down and don't respond properly.

jacklul:
>We can't ask user to edit the file with the token, maybe we can find another api for forecast that doesn't require the authentication. Is really not working anymore? Because sometimes the service seem down and don't respond properly.

MSN Weather?
(http://weather.service.msn.com/data.aspx?weasearchstr=Amsterdam,%20Netherlands&culture=EN&weadegreetype=C&src=php-script)

I've been using it for years in Rainmeter.

noplanman:
Like @MBoretto said, you can use the [command config](https://github.com/akalongman/php-telegram-bot#commands-configuration) to pass a user-specific variable.
You can check the [Date command](https://github.com/akalongman/php-telegram-bot/blob/master/src/Commands/UserCommands/DateCommand.php#L58) to see how to use it 👍

CrazyRedScorp:
Thanks!

PR #152, opened 2016-04-17
fix getMe() exception

jacklul:
Currently executing getMe request returns an exception (it tries to use message object for the returned data).


MBoretto:
Thanks!

PR #159, opened 2016-04-24
fix wrong parameters description

littleguga:
It works great with: 
```php
        $filename = __DIR__ . time() . ".jpg";
        file_put_contents($filename, fopen($link, 'r'));
        $result = Request::sendPhoto(
            [
                'chat_id' => Config::getTelegramChat(),
            ],
            $filename
        );
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -314,7 +314,7 @@ To send a local photo, provide the file path as the second parameter:
 
 ```php
 $data = ['chat_id' => $chat_id];
-$result = Request::sendPhoto($data, $telegram->getUploadPath() . '/image.jpg');
+$result = Request::sendPhoto($data, 'full-path-on-your-server-to/image.jpg');
 ```
 
 If you know the `file_id` of a previously uploaded file, just include it in the data array:


README.md
```diff
@@ -314,7 +314,7 @@ To send a local photo, provide the file path as the second parameter:
 
 ```php
 $data = ['chat_id' => $chat_id];
-$result = Request::sendPhoto($data, $telegram->getUploadPath() . '/image.jpg');
+$result = Request::sendPhoto($data, 'full-path-on-your-server-to/image.jpg');
```
MBoretto: you should load you location with:
https://github.com/akalongman/php-telegram-bot/blob/master/src/Telegram.php#L621

littleguga:
@MBoretto why I should this, when i already know the path to image and it works?
Also I'm not sure that ```$telegram->getUploadPath()``` returns the correct path(e.g. on Win it shows: ```D:\Development\Web\projects\vk-to-telegram-bot\vendor\longman\telegram-bot\src/../Upload```
I got his with: ```echo $telegram->getUploadPath();```

By the way, I suppose that I need to use ```setUploadPath($folder)``` to set folder various from ```\vendor\longman\telegram-bot\src```

MBoretto:
You are right for what concern the wrong path on windows. He're you can see the  automatic initialization intended for linux, we can open an issue for this. 
https://github.com/akalongman/php-telegram-bot/blob/master/src/Telegram.php#L160

The 'UploadPath()' is intended to have a know path where all the images to upload are stored, that's because can be used internally from the library. Of course you can use it also as you mention, but i would like to leave these information in the documentation.

littleguga:
@MBoretto thanks, now I understand meaning if this function. So I close PR.

Thanks for attention!

PR #163, opened 2016-04-25
Add a location button in the Survey example

juananpe:
Add a location button in a custom keyboard so the user can finish the survey example. 
Store the user's picture id in the conversation notes.

### DIFF
--- a/src/Commands/UserCommands/SurveyCommand.php
+++ b/src/Commands/UserCommands/SurveyCommand.php
@@ -151,8 +151,22 @@ class SurveyCommand extends UserCommand
                     $this->conversation->notes['state'] = 4;
                     $this->conversation->update();
 
+                    $keyboard = [];
+                    $keyboard[] = [
+                        [ 'text' => 'request_contact', 'request_contact' => true ],
+                        [ 'text' => 'request_location', 'request_location' => true ]
+                            ];
+                    $reply_keyboard_markup = new ReplyKeyboardMarkup(
+                        [
+                            'keyboard' => $keyboard ,
+                            'resize_keyboard' => true,
+                            'one_time_keyboard' => true,
+                            'selective' => true
+                            ]
+                    );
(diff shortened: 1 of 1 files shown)


juananpe:
I'd really appreciate to get some constructive feedback about this PR, even if you think that it should go directly to /dev/null , tell me why :)

noplanman:
Hi @juananpe 

First off, I really like the idea of the buttons!

Now for feedback:

- It would be nicer to have some human-readable text on the buttons, like "Select Contact" and "Current Location" or something like that.
- Regarding `request_contact`, that only sends the contact name and number. How can you add the location using that option? (We need a location object at this point!)
- Maybe the code could be made a bit more understandable by compressing it:
```php
$data['reply_markup'] = new ReplyKeyboardMarkup(
    [
        'keyboard' => [[
            [ 'text' => 'request_contact',  'request_contact'  => true ],
            [ 'text' => 'request_location', 'request_location' => true ],
        ]],
        'resize_keyboard'   => true,
        'one_time_keyboard' => true,
        'selective'         => true,
    ]
);
```
- Your PR needs to be pointed towards the `develop` branch, not `master`.

Hope that helps 😊

MBoretto:
Hello @juananpe , first of all thanks for contributing! 
I see your pr but just today i found some time to  try it.
Noplanman already pointed out some problems:
- When i select request contact  the survey stall because you need a  location object to access the next state. To overcome this  we can create an additional state where the bot ask the phone number with 'request_contact' button.
Let us know what you think. 
I'm really interested to include the new api features in this example!

juananpe:
Hi guys, thanks for the feedback. I totally agree with all of your comments:

* adding a "Select Contact" button when I'm requesting a location is somewhat weird, so following @MBoretto 's suggestion I'll add a new step in the survey, asking the user to share his/her phone number (and remove that button from the location request step)

* I'll point the PR towards the `develop` branch (instead of `master`)

PR #172, opened 2016-05-02
Add location and contact buttons to the Survey example

juananpe:
Add Location button in a custom keyboard so the user can finish the survey example.
Create an additional state where the bot asks the phone number with 'request_contact' button.
Store the user's picture id and user's phone number in the conversation notes.


juananpe:
After struggling with the branch model I think that I finally managed to contribute with a PR that is following the guidelines [commented in the previous PR](https://github.com/akalongman/php-telegram-bot/pull/163). If not, don't hesitate to inform me about it ;-P

MBoretto:
:+1:  will test asap

MBoretto:
I merged in develop after a small fix! Thanks for contributing!

PR #188, opened 2016-05-17
Misc fixes

jacklul:
Added missing things from bots 2.0 update.
Made some small visual changes to the code for better viewing.
Some small fixes.


noplanman:
The Entities will get a major overhaul at some point, but thanks a lot for going through the bulk and cleaning it up 😃 

@MBoretto and I have discussed about simplifying it a lot, to prevent all the duplicate and redundant code in all the files.

MBoretto:
@noplanman we should open an issue describing our idea, then we can assign it to someone ;)

noplanman:
Done #193

PR #189, opened 2016-05-17
chosen_inline_result table and make utf8mb4 default

jacklul:
Table for storing chosen inline results should be called chosen_inline_result, not chosen_inline_query, this is for consistency.

utf8mb4 will be the default encoding, user can override it by
```$telegram->enableMySQL($mysql_credentials, '', 'utf8');``` (second empty parameter is table_prefix)
I find this to be easier solution than introducing new variable in Telegram.php and adding setter and getter for it.

Moving to utf8mb4 as default is a must, without it we can't store emojis in tables and from my experience with my bots I see a lot of people are using emojis in names which results in DB full of empty values.


MBoretto:
I support moving to utf8mb4, do you consider utf8mb4_unicode_520_ci collation? It has advantages handling emoticons:
http://stackoverflow.com/questions/36633893/mariadb-select-wrong-unicode-character

ps: I use 520 collation in my project!  :)

jacklul:
Good suggestion, it seems to be better choice.

noplanman:
I noticed that travis doesn't know that collation, which is why the tests are failing:
`ERROR 1273 (HY000) at line 1 in file: 'structure.sql': Unknown collation: 'utf8mb4_unicode_520_ci'`

We'll need to find some solution for that.

MBoretto:
@noplanman @jacklul need to read hot to fix this with travis - ci.  Please guys wait to merge this in develop there are several that have to be addressed

MBoretto:
Ok so far so good.
I have a question:
When I introduced inline query i kept inline_query and chosen_inline_result in the same table. Could you explain what are the advantages of dividing those structures?   (I didn't use inline too much)

jacklul:
Purely to match the update object - chosen inline **result**, it makes more sense than chosen inline **query**

MBoretto:
Thanks!

PR #190, opened 2016-05-17
Quick Fix

jacklul:
Missed this in #188, sorry!

### DIFF
--- a/src/Entities/MessageEntity.php
+++ b/src/Entities/MessageEntity.php
@@ -18,23 +18,22 @@ class MessageEntity extends Entity
     protected $url;
 
     /**
-     * @todo check for value from this list: https://core.telegram.org/bots/api#messageentity
-     * @todo "empty($this->offset) && $this->offset != 0" - not an ideal solution?
+     * @todo check for type value from this list: https://core.telegram.org/bots/api#messageentity
      */
     public function __construct(array $data)
     {
         $this->type = isset($data['type']) ? $data['type'] : null;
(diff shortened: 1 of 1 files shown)


src/Entities/MessageEntity.php
```diff
@@ -18,23 +18,22 @@ class MessageEntity extends Entity
...
     {
         $this->type = isset($data['type']) ? $data['type'] : null;
-        if (empty($this->type)) {                               //
+        if (empty($this->type)) {
             throw new TelegramException('type is empty!');
         }
 
         $this->offset = isset($data['offset']) ? $data['offset'] : null;
-        if (empty($this->offset) && $this->offset != 0) {       //
+        if ($this->offset === '') {
```
noplanman: I haven't checked, but you're sure that `$this->offset` can't be `0`?

jacklul: Straight from logs from one of my bots:
```
"entities":[{"type":"bot_command","offset":0,"length":7}]
"entities":[{"type":"bold","offset":350,"length":0}]
```

noplanman: Ok, so it can be `0`. And clearly you've tested this, yes? 😊 

Sorry, I'm sometimes a bit paranoid about changes like this, as they sometimes might introduce the little edge cases that produce weird errors.

jacklul: Yep, tested!

MBoretto:
We can create a test for this

PR #200, opened 2016-05-23
Adventures with Monolog

MBoretto:
Would like to know your feedback about monolog integration.
The class TelegramLog is made with the singleton pattern in mind.
Don't want to describe too much here because i wrote 'everything' in the doc!

Things done:
- Introducing Monolog with TelegramLog.php class
- Update examples of hook.php and getUpdates.php
- Introducing doc dir
- Introducing utils/importFromLog.php that let you refill the database from 0

Let me know! @akalongman @noplanman @jacklul


src/Request.php
```diff
@@ -10,6 +10,7 @@
 
 namespace Longman\TelegramBot;
 
+use Longman\TelegramBot\TelegramBot;
```
noplanman: I assume you mean `use Longman\TelegramBot\TelegramLog` here, right?
If so, you can remove this line completely, as `TelegramLog` is on the same namespace level.

noplanman:
@MBoretto **Nice work!!**

I've made a mini-PR [here](https://github.com/MBoretto/php-telegram-bot/pull/11) with some changes.

**My feedback and questions:**
- Making the `TelegramLog` class work like a singleton is a great solution!
- Why not use the `TelegramLog` static methods directly instead of adding an extra layer of methods to `Telegram`? (Namely `setErrorLog`, `setDebugLog` and `setUpdateLog`) They're doing exactly that, simply calling the static method. I think it's better to not include it in `Telegram`, because it makes that class bigger and doesn't add any extra value. Feel free to explain your view and thoughts, these are just mine! 😊 
- You need to make sure that a valid path has been passed to the [`init...` methods](https://github.com/akalongman/php-telegram-bot/pull/200/files#diff-c686a02e33036e126be940fad368ab7bR85), otherwise it silently dies. (I haven't added this to the PR)
- The `error()` method is never called. Shouldn't the `TelegramException` class use it? Or is the idea that the user must manually add it to the `try..catch`?

MBoretto:
- Yes we can use directly the `TelegramLog` static methods without all the stuff in `Telegram.php`. I don't see any disadvantage!  
- Yes you are right for what concern the control in the initialization. I wonder which is the best way to handle this, I mean how can handle error of the logging manager?? Maybe I can throw an exception (not the Telegram exception one). Last question (maybe stupid) should a static initialize always return void? (I work on some code that always return void from a `static ::initialize()`)
- `error()` is called here:   
https://github.com/MBoretto/php-telegram-bot/blob/monolog/examples/hook.php#L62  
https://github.com/MBoretto/php-telegram-bot/blob/monolog/examples/getUpdatesCLI.php#L71  

now i added them to the pr! ;) thanks

MBoretto:
New commit with @noplanman  suggestion.

noplanman:
> should a static initialize always return void?

I don't know. The main advantage I can think of, is to discourage method chaining (if for example the object itself is returned) when the initialisation doesn't complete successfully.
Also, the initialisation is a 1-time operation, so I guess it's OK to give it it's own line of code when called 😊

src/TelegramLog.php
```diff
@@ -10,6 +10,7 @@
 
 use Monolog\Logger;
 use Monolog\Handler\StreamHandler;
+use Longman\TelegramBot\Exception\TelegramLogException;
 
 namespace Longman\TelegramBot;
```
noplanman: `namespace` declaration must be before the `use` ones.
Otherwise the `TelegramLogException` can't be found, because it's searched for as `Longman\TelegramBot\TelegramLogException` (missing the `\Exception`)

src/TelegramLog.php
```diff
@@ -84,6 +85,9 @@ public static function initialize(\Monolog\Logger $external_monolog = null)
      */
     public static function initErrorLog($path)
     {
+        if (empty($path)) {
```
noplanman: When comparing strings, it's better to compare `=== ''` or `!== ''`, because maybe I want my filename to be `0` (which is regarded as empty)
I noticed that this bug is in Monolog too... When I call the file `1` it works, `0` doesn't, because it probably also has an `empty()` check somewhere.

noplanman:
From my side, all ago for merging into develop! 👍 

@jacklul @akalongman You ok with that?

PR #210, opened 2016-05-30
Some structure.sql fixes

Sogl:
1. Duplicate table `edited_message` removed
2. In MySQL 5.7 with default mode `STRICT_TRANS_TABLES` you can see error: `SQL Error 1101: BLOB/TEXT/GEOMETRY/JSON column '' can't have a default value`. To avoid this I removed all default values from TEXT columns.

Tested and works on MySQL 5.7.


Sogl:
Screen of error in HeidiSQL for 2nd point:
errr

noplanman:
Thanks @Sogl 
I'll close this PR here as it's pointing to the master branch.

Please **point your PR to develop branch**, so that we can then merge it!

PR #212, opened 2016-05-31
Add inline line breaks in logs.

noplanman:
By default, Monolog removes all line breaks, resulting in a log file that has really long lines.
For debug purposes, it's very difficult to find the relevant info in a long line.

This PR enables line breaks in debug and error logs.

### DIFF
--- a/src/TelegramLog.php
+++ b/src/TelegramLog.php
@@ -100,7 +100,11 @@ class TelegramLog
         }
         self::initialize();
         self::$error_log_path = $path;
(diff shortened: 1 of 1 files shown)


MBoretto:
Thanks for pointing out this, gonna merge! I wonder if this line breaks also involve the telegram raw log?

noplanman:
If by raw you mean the JSON update messages, then no, because there are no line breaks in those, they're just 1 line each.

PR #213, opened 2016-06-01
Guzzleify

### LINKED ISSUE #168: Introduce Guzzle
We would like to make the request system easier, we are thinking to switch to guzzle.
Any comments?
Marco & Armando (first php-telegram-bot meeting!! @akalongman @jacklul  )

noplanman:
This PR introduces Guzzle to replace cURL calls.

Apart from the main `Request` class, the `/date` and `/weather` commands have been updated / rewritten too.

@jacklul You mentioned in the issue that async requests could be used, how did you envision this?
I guess I would need to add extra methods to allow the configuration of the Guzzle Client, right?

closes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,10 +464,14 @@ Inside *examples/Commands/* there are some samples that show how to use types.
 
 #### Commands Configuration
 
-With this method you can set some command specific parameters, for
-example, google geocode/timezone api key for date command:
+With this method you can set some command specific parameters:
+
 ```php
+//Google geocode/timezone API key for date command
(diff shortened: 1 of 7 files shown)


jacklul:
As far as I know you can also make requests and do not wait for result which will speed up the process when you make a lot of requests to telegram API, for example - sending mass notifications.

_Also in documentation there is a mention about answering webhooks directly by outputing the reply with 200 status and Content-Type: application/json and json string in page content, which seems like the fastest way to serve webhooks but only if always reply with one request to their API, This could be implemented someday..._

noplanman:
Argh, make a new PR, will continue discussion there: #214

PR #214, opened 2016-06-02
Guzzleify

### LINKED ISSUE #168: Introduce Guzzle

noplanman:
This (corrected) PR introduces Guzzle to replace cURL calls.

Apart from the main `Request` class, the `/date` and `/weather` commands have been updated / rewritten too.

closes #168 
(old PR with some comments: #213)

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,10 +464,14 @@ Inside *examples/Commands/* there are some samples that show how to use types.
 
 #### Commands Configuration
 
-With this method you can set some command specific parameters, for
-example, google geocode/timezone api key for date command:
+With this method you can set some command specific parameters:
+
 ```php
+//Google geocode/timezone API key for date command
 $telegram->setCommandConfig('date', ['google_api_key' => 'your_google_api_key_here']);
+
+//OpenWeatherMap API key for weather command
+$telegram->setCommandConfig('weather', ['owm_api_key' => 'your_owm_api_key_here']);
 ```
 
 ### Admin Commands
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
         "php": ">=5.5.0",
         "ext-pdo": "*",
         "ext-curl": "*",
-        "monolog/monolog": "^1.19"
+        "monolog/monolog": "^1.19",
+        "guzzlehttp/guzzle": "~6.0"
     },
     "autoload": {
         "psr-4": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,9 +4,180 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "hash": "49fc795b5cdb1b689416910c6bed4023",
-    "content-hash": "b0ad7b5aff6fea3eb03b195779708bda",
+    "hash": "3d67bf3fdb85b051d8f76a6098654143",
+    "content-hash": "e896da909d596879e7fcf1a2b7898113",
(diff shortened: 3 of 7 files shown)


noplanman:
(quoting @jacklul from old PR)
> As far as I know you can also make requests and do not wait for result which will speed up the process when you make a lot of requests to telegram API, for example - sending mass notifications.
Additional method could be added like - Request:setAsync(true), calling it before API request would make it not wait for result. Something like that.
_Also in documentation there is a mention about answering webhooks directly by outputing the reply with 200 status and Content-Type: application/json and json string in page content, which seems like the fastest way to serve webhooks but only if always reply with one request to their API, This could be implemented someday..._

Could you link the page of the documentation that mentions the faster method of processing the webhook?

Also, about the async, I see your point regarding mass notifications. Does the API have a limit there? We'd have to make sure not to pass the limit, or some notifications might not get through at all.

jacklul:
Yes, there are limits: https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this
I think there are also general limits that also apply for clients like sending too many messages in short time.

https://core.telegram.org/bots/api#making-requests-when-getting-updates
https://core.telegram.org/bots/faq#how-can-i-make-requests-in-response-to-updates

One of my bots is using this, answers directly by outputing this to webhook requests:
```{"method":"answerCallbackQuery","callback_query_id":$id,"text":"Too many requests!","show_alert":true}```
even before being handled by the library (handled by a bash cgi script)

**Comment by noplanman** [comment]
@jacklul Makes sense to implement this into the library! 👍 

The library needs to be run anyway though, to determine if the passed message is a command / generic message etc.

The way I see it, there aren't any downsides in implementing this for Webhook updates are there?

**Comment by noplanman** [comment]
Right, I've been exploring a bit and found a few issues:

Using the direct API responses only works for a single command, not multiple ones. So all commands that perform multiple actions (`/whoami` for example), only the first one goes through. I've messaged Telegram SupportBot to ask if there is a way of sending multiple responses in one go, let's see.

When using Guzzle, the keyboards don't work any more. This is due to the way Guzzle sets up the HTTP query [here](https://github.com/guzzle/guzzle/blob/6.2.0/src/Client.php#L294). They use [`http_build_query`](https://secure.php.net/manual/en/function.http-build-query.php), which takes an array or object as a parameter. The problem is that the keyboards are objects, so the `http_build_query` gets the public variables of that object to map the query, instead of just casting it to a string. So instead of the `__toString()` method being called on the object, an empty string is returned, because the object has no public member variables. As far as I can see, the only way of fixing this, is to either cast the keyboards to string when adding them to the data array being passed to the request...
```php
$data['reply_markup'] = (string)$reply_keyboard_markup;
```
**instead of just**
```php
$data['reply_markup'] = $reply_keyboard_markup;
```
...or alternatively, which @MBoretto and I had discussed already when we met up, to create a keyboard helper class that returns a string in the first place without creating any `XKeyboardY` object.
Whichever path we choose (I highly encourage the latter to make keyboard creation easier!!), all custom commands that our users may have that require keyboards, need to be updated!

noplanman:
[e66eb19](https://github.com/akalongman/php-telegram-bot/pull/214/commits/e66eb190fdccf1d54639bec917b472e8dc23d6b6) fixes the problem with the Keyboard objects.

It still needs to be addressed though❗

PR #219, opened 2016-06-08
Fix cancel command if no database is connected.

noplanman:
Message mustn't have an empty text, otherwise Telegram API is sad 😢 

So just output a simple message instead, since there is nothing to cancel if there is no database connection, as no conversation can be active anyway.


noplanman:
@akalongman @MBoretto @jacklul 

You guys ok with me merging this?

PR #226, opened 2016-06-15
Small optimization

jacklul:
instead of inserting whole message into callback query entry reference it in database 
(inserts it to the message table)

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -622,8 +622,7 @@ class DB
                 $message_id = $message->getMessageId();
 
                 $sth = self::$pdo->prepare('SELECT * FROM `' . TB_MESSAGE . '`
-                    WHERE `id` = ' . $message_id . ' AND `chat_id` = ' . $chat_id . ' LIMIT 1'
-                );
+                    WHERE `id` = ' . $message_id . ' AND `chat_id` = ' . $chat_id . ' LIMIT 1');
                 $sth->execute();
 
                 if ($sth->rowCount() > 0) {
--- a/src/DB.php
+++ b/src/DB.php
@@ -569,7 +569,7 @@ class DB
             $sth_insert_chosen_inline_result->bindParam(':result_id', $result_id, \PDO::PARAM_STR);
             $sth_insert_chosen_inline_result->bindParam(':user_id', $user_id, \PDO::PARAM_INT);
             $sth_insert_chosen_inline_result->bindParam(':location', $location, \PDO::PARAM_INT);
-            $sth_insert_chosen_inline_result->bindParam(':inline_message_id', $inline_message_id, \PDO::PARAM_INT);
+            $sth_insert_chosen_inline_result->bindParam(':inline_message_id', $inline_message_id, \PDO::PARAM_STR);
             $sth_insert_chosen_inline_result->bindParam(':query', $query, \PDO::PARAM_STR);
             $sth_insert_chosen_inline_result->bindParam(':created_at', $date, \PDO::PARAM_STR);
 
@@ -596,10 +596,10 @@ class DB
         try {
             $mysql_query = 'INSERT IGNORE INTO `' . TB_CALLBACK_QUERY . '`
                 (
-                `id`, `user_id`, `message`, `inline_message_id`, `data`, `created_at`
+                `id`, `user_id`, `message_id`, `inline_message_id`, `data`, `created_at`
                 )
                 VALUES (
-                :callback_query_id, :user_id, :message, :inline_message_id, :data, :created_at
+                :callback_query_id, :user_id, :message_id, :inline_message_id, :data, :created_at
                 )';
 
             $sth_insert_callback_query = self::$pdo->prepare($mysql_query);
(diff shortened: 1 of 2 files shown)


src/DB.php
```diff
@@ -615,12 +615,21 @@ public static function insertCallbackQueryRequest(CallbackQuery &$callback_query
             }
 
             $message = $callback_query->getMessage();
+            $chat_id = null;
+            $message_id = null;
+            if ($message) {
+                $chat_id = $callback_query->getMessage()->getChat()->getId();
+                $message_id = $callback_query->getMessage()->getMessageId();
```
noplanman: You can use `$message` from above here.

jacklul: Oh right

src/DB.php
```diff
@@ -660,6 +660,12 @@ public static function insertMessageRequest(Message &$message)
         $chat_id = $chat->getId();
 
         $date = self::getTimestamp($message->getDate());
+        $edit_date = self::getTimestamp($message->getEditDate());
+
+        if ($edit_date == $date) {
```
noplanman: This condition will (pretty much) never be true, because `edit_date` is never set at this point. So `$edit_date` will always have the current timestamp (because `null` is being passed to `getTimestamp()`).
Hence, `$date` is always a bit earlier than `$edit_date`.

What you'd want is something like this:
```php
$date = self::getTimestamp($message->getDate());
if ($edit_date = $message->getEditDate()) {
    $edit_date = self::getTimestamp($edit_date);
    if ($date === $edit_date) {
        $edit_date = null;
    }
}
```

But like I said, this always fails, because if a message is being edited, `insertEditedMessageRequest()` is called, not `insertMessageRequest()`

Please correct me if I'm wrong!

src/DB.php
```diff
@@ -758,6 +766,7 @@ public static function insertMessageRequest(Message &$message)
             $sth->bindParam(':message_id', $message_id, \PDO::PARAM_INT);
             $sth->bindParam(':user_id', $from_id, \PDO::PARAM_INT);
             $sth->bindParam(':date', $date, \PDO::PARAM_STR);
+            $sth->bindParam(':edit_date', $edit_date, \PDO::PARAM_STR);
```
noplanman: This is so that the latest edit date is saved with the original message, right?
Unfortunately this doesn't get called when a message is edited, because `insertEditedMessageRequest()` is called, not `insertMessageRequest()`

Please correct me if I'm wrong!

jacklul: You're right about the ```$edit_date```, will adjust code to your suggestion.

```insertEditedMessageRequest()``` currently applies only to edited messages by the user

The changes I made in the code are specifically to update bot's message contaning callback query inline buttons (full message object is sent with the callback query)
edited messages that were pasted from inline mode are using ```insertEditedMessageRequest()``` (they are in fact posted by user - 'xxx via superbot' or something like that)

At first I wanted to insert every message edited by the bot with ```insertEditedMessageRequest()``` but couldn't find a way to do that, any ideas?

Comment (noplanman) [inline src/DB.php]: Mind posting a super-simple command that uses the callback query you're talking about please?
Haven't looked into those that much yet and it would make my life a lot easier 😊 
Then I'll understand exactly what you mean and we can find a nice solution!

Comment (jacklul) [inline src/DB.php]: I believe current CallbackqueryCommand is good example.

As per documentation https://core.telegram.org/bots/api#callbackquery callback query will either have Message object (messages posted by bot) [case 1] or inline_message_id (inline results posted by user via bot) [case 2]

Case 1:
- imagine /settings command that uses inlline buttons to navigate inside same message via editing it
- if this message is not in database it should be inserted to ```message``` table (ideally user's first interaction will insert the first state of the message - original message that was sent by the bot)
- every time message is edited (user navigating the settings buttons) either update the message in ```message``` table or add edited message to ```edited_message``` (second solution makes more sense and can provide edit history so people could see in their project in what order people are navigating settings therefore knowing which options catch bigger interest)

Case 2: (does not need any code changes)
- imagine the bot is providing simple images from inline search
- user posts result from the inline search with thumb up and down buttons
- clicking those buttons changes numbers next to each voting option (thumb up/down) which results in the message being edited by user (edited_message Update object is sent)

I'm currently using code solution that adds another query to be executed with each callback query, not an ideal solution I must say:

```php
public static function insertCallbackQueryRequest(CallbackQuery &$callback_query)
    {
        if (!self::isDbConnected()) {
            return false;
        }

        try {
            $mysql_query = 'INSERT IGNORE INTO `' . TB_CALLBACK_QUERY . '`
                (
                `id`, `user_id`, `chat_id`, `message_id`, `inline_message_id`, `data`, `created_at`
                )
                VALUES (
                :callback_query_id, :user_id, :chat_id, :message_id, :inline_message_id, :data, :created_at
                )';

            $sth_insert_callback_query = self::$pdo->prepare($mysql_query);

            $date = self::getTimestamp(time());

            $callback_query_id = $callback_query->getId();
            $from = $callback_query->getFrom();
            $user_id = null;
            if (is_object($from)) {
                $user_id = $from->getId();
                self::insertUser($from, $date);
            }

            $message = $callback_query->getMessage();
            $chat_id = null;
            $message_id = null;
            if ($message) {
                $message_id = $message->getMessageId();

                $sth = self::$pdo->prepare('SELECT * FROM `' . TB_MESSAGE . '`
                    WHERE `id` = ' . $message_id . ' AND `chat_id` = ' . $chat_id
                );
                $result = $sth->execute();

                if ($result) {
                    if ($sth->rowCount() > 0) {
                        self::insertEditedMessageRequest($message);
                    } else {
                        self::insertMessageRequest($message);
                    }
                }
            }

            $inline_message_id = $callback_query->getInlineMessageId();
            $data = $callback_query->getData();

            $sth_insert_callback_query->bindParam(':callback_query_id', $callback_query_id, \PDO::PARAM_INT);
            $sth_insert_callback_query->bindParam(':user_id', $user_id, \PDO::PARAM_INT);
            $sth_insert_callback_query->bindParam(':chat_id', $chat_id, \PDO::PARAM_INT);
            $sth_insert_callback_query->bindParam(':message_id', $message_id, \PDO::PARAM_INT);
            $sth_insert_callback_query->bindParam(':inline_message_id', $inline_message_id, \PDO::PARAM_STR);
            $sth_insert_callback_query->bindParam(':data', $data, \PDO::PARAM_STR);
            $sth_insert_callback_query->bindParam(':created_at', $date, \PDO::PARAM_STR);

            return $sth_insert_callback_query->execute();
        } catch (PDOException $e) {
            throw new TelegramException($e->getMessage());
        }
    }
```

I basically check if the original message exists in ```message``` table and if not insert it there, otherwise insert into ```edited_message``` and reference original.

This code is a replacement for my 'ON DUPLICATE' solution in ```insertMessageRequest()``` commit.

MBoretto:
@jacklul Great! Thank you! I would like to test it a while then I will merge!

PR #234, opened 2016-06-26
inline query command fixed

MBoretto:
Found a little bug..


noplanman:
@MBoretto, not a bug, a feature 😉 

The line is correct as it was, since we're defining the `$data` array that doesn't exist yet.
Your change converts a non-existing variable into an array.

Is the command causing you problems? If so, we need to look closer why.

MBoretto:
another feature!

PR #236, opened 2016-06-29
various

MBoretto:
Doc block, various error, fixed thanks to incredible new tools..

### DIFF
--- a/src/Commands/AdminCommands/SendtochannelCommand.php
+++ b/src/Commands/AdminCommands/SendtochannelCommand.php
@@ -33,7 +33,7 @@ class SendtochannelCommand extends AdminCommand
     /**
      * Conversation Object
      *
-     * @var Longman\TelegramBot\Conversation
+     * @var \Longman\TelegramBot\Conversation
      */
     protected $conversation;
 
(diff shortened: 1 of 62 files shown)


noplanman:
Good to merge, mostly DocBlock changes and some code changes that don't affect any behaviour.

👍

src/Commands/AdminCommands/SendtochannelCommand.php
```diff
@@ -33,7 +33,7 @@ class SendtochannelCommand extends AdminCommand
     /**
      * Conversation Object
      *
-     * @var Longman\TelegramBot\Conversation
+     * @var \Longman\TelegramBot\Conversation
```
noplanman: @MBoretto Did you deliberately leave all full namespaces as opposed to simplified ones?

PhpStorm automatically prefers simplified ones for me.

@akalongman @jacklul 
Asking you guys too, which standard shall we go by?
```php
@var \Longman\TelegramBot\Conversation
```
or
```php
@var Conversation
```
(this obviously implies that we have a `use Longman\TelegramBot\Conversation` at the top.

MBoretto:
I add it because with the \ in the front and the full namespace php storm seems able to recognise the type of the variable

noplanman:
Interesting, it seems to work for me without the whole namespace if there is an appropriate `use` present.
So you would suggest to add the full namespace to **all** docblock variables?

PR #238, opened 2016-07-03
Statistics command

Excody:
Statistics command


jacklul:
Why not including both PR as one? Correct travis errors in #239

Also: getCountIncomingCommands will look better than getCountIncomeCommands (same with getCountIncomeMessageAll)

MBoretto:
@Excody waiting for new PR!

PR #239, opened 2016-07-03
Add new methods

Excody:
Add new methods for statistic command


jacklul:
See https://github.com/akalongman/php-telegram-bot/pull/238#issuecomment-230151347

PR #240, opened 2016-07-03
Add flexible methods

### LINKED ISSUE #230: More flexible methods
It would be great to have some more flexible methods regarding the parameters that they allow.

e.g.
```php
// Now
$telegram->enableAdmins(123); // fails, can't assign multiple admins in 1 call
$telegram->enableAdmins([123, 456]); // works, must be an array!

$telegram->addCommandsPath('/some/commands') // works, must be a string
$telegram->addCommandsPath(['/some/commands', '/other/commands']) // fails, can't add multiple paths in 1 call

// Then
$telegram->enableAdmins(123); // works, add a single one
$telegram->enableAdmins([123, 456]); // works, add multiple ones

$telegram->addCommandsPath('/some/commands') // works, add a single one
$telegram->addCommandsPath(['/some/commands', '/other/commands']) // works, add multiple ones
```

The same "could" apply for the `setCommandConfig` method.

As you can see in the above example, we have `enableAdmins` (plural) and `addCommandsPath` (singular), which makes sense for how it is now.

There are 2 ways to adapt the code:
- Create the extra methods `enableAdmin` (singular) and `addCommandsPaths` (plural)
- Create just the plural ones and remove the singular ones (or rather, set them as *deprecated*)

I think the first is easier for users to understand and having a few extra methods shouldn't be a problem.

What are your views on this?

noplanman:
Add extra methods for enabling admins and adding custom commands paths.
Closes #230

### DIFF
--- a/tests/Unit/TelegramTest.php
+++ b/tests/Unit/TelegramTest.php
@@ -27,11 +27,36 @@ class TelegramTest extends TestCase
     private $telegram;
 
     /**
+     * @var array
+     */
+    private $custom_commands_paths = [
+        '/tmp/php-telegram-bot-custom-commands-1',
+        '/tmp/php-telegram-bot-custom-commands-2',
+        '/tmp/php-telegram-bot-custom-commands-3',
+    ];
+
+    /**
     * setUp
     */
     protected function setUp()
     {
         $this->telegram = new Telegram('testapikey', 'testbotname');
+
+        // Create a few custom commands paths.
+        foreach ($this->custom_commands_paths as $custom_path) {
+            mkdir($custom_path);
+        }
+    }
+
+    /**
+     * tearDown
+     */
+    protected function tearDown()
+    {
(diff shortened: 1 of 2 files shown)


tests/Unit/TelegramTest.php
```diff
@@ -71,6 +96,53 @@ public function getBotName()
...
+        $this->assertCount(3, $tg->getAdminList());
+
+        $tg->enableAdmin('a string?');
+        $this->assertCount(3, $tg->getAdminList());
+    }
+
+    /**
+     * @test
+     */
+    public function addCustomCommandsPaths()
```
MBoretto: Should test function start with test in the front?

noplanman: It's nicer, yes, because then it wouldn't need the `@test` comment above each function.

But I thought of changing just that for all tests in a single PR, OK?

PR #241, opened 2016-07-03
Return an empty response when an empty inline query is passed.

noplanman:
Prevent an error exception when an empty inline query is being handled.


jacklul:
Wait, I think Telegram expects AnswerInlineQuery even to empty queries, I think with your solution user will see infinity loading icon with empty query!

noplanman:
Aaah yes, you're right!
So we'll need to find a different solution. The problem is, that if the inline query is empty, I get the following error:
`{"ok":false,"error_code":400,"description":"Bad Request: Message must be non-empty"}`

jacklul:
Maybe just 
`'input_message_content' => new InputTextMessageContent([ 'message_text' => 'you enter: ' . $query
`

PR #248, opened 2016-07-10
allow value 0 for duration

MBoretto:


src/Entities/Audio.php
```diff
@@ -34,7 +34,7 @@ public function __construct(array $data)
         }
 
         $this->duration = isset($data['duration']) ? $data['duration'] : null;
-        if (empty($this->duration)) {
+        if ($this->duration === '') {
```
noplanman: Also check for `null`, in case the field doesn't exist at all.
```php
if ($this->duration === null || $this->duration === '')
```

I'm wondering if this is ever the case if the update comes from Telegram.
Creating an own object, this could obviously happen...

PR #254, opened 2016-07-16
Fixes issue with file uploading

Livich:
The problem is that there is an invalid argument for Guzzle's `post` method. Bot can't upload files to users using `form_params`, so this fix makes bot to use `multipart` key.


noplanman:
I've finally tested it and it works perfectly!

Thanks so much @Livich 👍

PR #257, opened 2016-07-22
Improved InlineKeyboardButton

iamraccoon:
Fixed TODO in InlineKeyboardButton:
check if only one of 'url, callback_data, switch_inline_query' fields is set, documentation states that only one of these can be used

### DIFF
--- a/src/Entities/InlineKeyboardButton.php
+++ b/src/Entities/InlineKeyboardButton.php
@@ -18,14 +18,13 @@ class InlineKeyboardButton extends Entity
     protected $url;
     protected $callback_data;
     protected $switch_inline_query;
-
+    private $hasOption = false;
 
     /**
      * InlineKeyboardButton constructor.
      *
-     * @todo check if only one of 'url, callback_data, switch_inline_query' fields is set, documentation states that only one of these can be used
-     *
      * @param array $data
+     * @throws TelegramException
      */
     public function __construct($data = array())
     {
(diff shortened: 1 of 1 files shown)


noplanman:
Hi @iamraccoon thanks for your PR!

For future PRs, please have a look at how the existing coding style is and try to adapt to it.
e.g. variables in `$snake_case`, not `$camelCase`, using `&&` instead of `and`, etc.
Keeps the code nice and clean, which is VERY important 👍 

I really appreciate your PR and the fix, but would suggest the following to make the code more readable and reliable:

Instead of adding an extra private variable, it would be cleaner to use `elseif` statements like so:
```php
if (!empty($data['url'])) {
    $this->url = $data['url'];
} elseif (!empty($data['callback_data'])) {
    $this->callback_data = $data['callback_data'];
} elseif (!empty($data['switch_inline_query'])) {
    $this->switch_inline_query = $data['switch_inline_query'];
} else {
    throw new TelegramException('You must use at least one of these fields: url, callback_data, switch_inline_query!');
}
```
or opt for an even more elegant solution using `foreach` like so:
```php
foreach (['url', 'callback_data', 'switch_inline_query'] as $param) {
    if (!empty($data[$param])) {
        $this->$param = $data[$param];
        return;
    }
}
throw new TelegramException('You must use at least one of these fields: url, callback_data, switch_inline_query!');
```

Also, no need to have `isset($var) && !empty($var)` in the same statement, just use `!empty($var)`, as the value needs to exist anyway to not be empty, if you know what I mean.
It's like writing `if ($var !== null && ! ($var === null))`, which is the same thing.

I hope you could learn something and see the value in my suggestions.

jacklul:
I think it would be better to throw an exception when developer uses more than one field.

noplanman:
@jacklul Agreed.
```php
$num_params = 0;
if (!empty($data['url'])) {
    $this->url = $data['url'];
    $num_params++;
}
if (!empty($data['callback_data'])) {
    $this->callback_data = $data['callback_data'];
    $num_params++;
}
if (!empty($data['switch_inline_query'])) {
    $this->switch_inline_query = $data['switch_inline_query'];
    $num_params++;
}
if ($num_params !== 1) {
    throw new TelegramException('You must use only one of these fields: url, callback_data, switch_inline_query!');
}
```
or respectively
```php
$num_params = 0;
foreach (['url', 'callback_data', 'switch_inline_query'] as $param) {
    if (!empty($data[$param])) {
        $this->$param = $data[$param];
        $num_params++;
    }
}
if ($num_params !== 1) {
    throw new TelegramException('You must use only one of these fields: url, callback_data, switch_inline_query!');
}
```

noplanman:
Replaced by #273

PR #258, opened 2016-07-23
Update structure.sql

aleksraiden:
Change table and collate encoding to utf16 - for correct save and display emdji (e.g. in Telegram game we using custom keyboard with text and emoji icon)

### DIFF
--- a/structure.sql
+++ b/structure.sql
@@ -8,7 +8,7 @@ CREATE TABLE IF NOT EXISTS `user` (
 
   PRIMARY KEY (`id`),
   KEY `username` (`username`)
-) ENGINE=InnoDB DEFAULT CHARSET=utf16_unicode_520_ci COLLATE=utf16_unicode_520_ci;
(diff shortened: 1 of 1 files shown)


noplanman:
utf8mb4 already supports emoji.
If I'm not mistaken, changing to UTF16 won't add any advantage and will increase the database size for every value saved.

UTF8 is variable 1 to 4 bytes.
UTF16 is variable 2 or 4 bytes.
UTF32 is fixed 4 bytes.
See here: http://stackoverflow.com/a/496340

@jacklul @MBoretto Am I correct here or have I missed something?

MBoretto:
That's the first time that I hear about this encoding, we switch to utf8mb4 some months ago to handle all the possible characters.  @aleksraiden which characters are missing? 
First results found on google: 
https://dev.mysql.com/doc/refman/5.5/en/charset-unicode-utf16.html

MBoretto:
So? I don't see any advantage to move to this encoding, would like to close this. @noplanman @jacklul ?

PR #260, opened 2016-07-30
fix enableAdmins()

jacklul:
Basically for me ```$telegram->enableAdmins(['xxxxxxxxxx']);``` doesn't work, admins_list array remains empty, it returns ```'Invalid value xxxxxxxxxx for admin.'``` error instead.

It seems like on my webhost my id isn't considered int, some systems might limit max int value and I guess thats what happens in my case... it's only weird that my webhosts limits it at length = 10.
I guess is_numeric + !is_float combo is good replacement here?

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -467,7 +467,7 @@ class Telegram
      */
     public function enableAdmin($admin_id)
     {
-        if (is_int($admin_id) && $admin_id > 0 && !in_array($admin_id, $this->admins_list)) {
+        if (is_numeric($admin_id) && !is_float($admin_id) && $admin_id > 0 && !in_array($admin_id, $this->admins_list)) {
             $this->admins_list[] = $admin_id;
         } else {
             TelegramLog::error('Invalid value "' . $admin_id . '" for admin.');


noplanman:
No need for `is_float`, they're part of the `is_numeric` check 👍

jacklul:
Without !is_float check it lets people use float numbers here...

noplanman:
Ach, silly me, didn't realize that it was a `! is_float`...

*Move along, nothing to see here* 😁

jacklul:
Nevermind, seems like ```$telegram->enableAdmins(['your_telegram_id']);``` causes this.
Readme and examples need updating i guess.

MBoretto:
Can we cast to string before check? (int)

noplanman:
@jacklul I just realised that my changes already **require** the parameter to be an int. Sorry about that, had it in my head differently.
So obviously, if the id is bigger than `PHP_INT_MAX`, `is_int` fails.
Using `is_numeric` would allow for bigger ones and strings that look like numbers (e.g `"5", "9.2E+18", etc.`), which we also don't want.

So the solution is to update the examples and the readme.

For the future, if there are breaking updates like this one, we should have a way to let the bot maintainers know, so that they can prepare for a coming release.
Any ideas on that?

noplanman:
The other thing I was thinking of: It would be cool if the `@username` could be used instead of the id. Would make it easier for bot maintainers to add the ids of other users, instead of having to check the database to find out the users id.

What do you think about this idea?

noplanman:
#261 is PR to update the examples and the readme. I've also added an extra test regarding numbers passed as string.

jacklul:
>The other thing I was thinking of: It would be cool if the `@username` could be used instead of the id. Would make it easier for bot maintainers to add the ids of other users, instead of having to check the database to find out the users id.

I was thinking about settings database, if someone set @username as admin bot could resolve it manually on the first user occurence and cache it in database

MBoretto:
I think that technically can be used the username instead if the id. But username can change so can identify a user as long as he doesn't change his username. Of course you can't change id!

noplanman:
@MBoretto, so true, hadn't thought of that.

PR #264, opened 2016-08-02
Update inline comments for hook.php and getUpdatesCLI.php

noplanman:
Make sure that logging is set directly after Telegram object initialisation. (thanks @jacklul for pointing this out)
Proofread and update utils.md file.

### DIFF
--- a/examples/hook.php
+++ b/examples/hook.php
@@ -1,16 +1,24 @@
 <?php
-//README
-//This configuration file is intended to run the bot with the webhook method.
-//Uncommented parameters must be filled
-//Please notice that if you open this file with your browser you'll get the "Input is empty!" Exception.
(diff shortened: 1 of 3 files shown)


examples/hook.php
```diff
@@ -22,46 +30,48 @@
...
     //$telegram->enableBotan('your_token');
 
     // Handle telegram webhook request
     $telegram->handle();
 } catch (Longman\TelegramBot\Exception\TelegramException $e) {
-    // Silence is gold!
-    // echo $e;
-    // log telegram errors
-    \Longman\TelegramBot\TelegramLog::error($e);
+    // Silence is golden!
```
MBoretto: ahahah
http://forum.wordreference.com/threads/silence-is-gold-or-golden.719876/

PR #289, opened 2016-08-28
fix constraint error

jacklul:
In case user spams inline keyboard button it will create a duplicated query which will produce constraint error, adding 'IGNORE' to the query shall fix this, it was general mistake of mine to not add it in the first place because other queries have eithier IGNORE or KEY UPDATE in the queries.


noplanman:
Has been added in #288

PR #291, opened 2016-09-02
Refactor entities

noplanman:
- Complete rewrite of the Entities, making them more lightweight and almost code-free.
- Methods are now defined by DocBlock comments instead of actual code.
- Magic getters and setters are used for all Entities, including InlineQueryResult objects.
- Updated tests and example code.
- New namespaces for Keyboards and InlineQuery entities.
- New Keyboard and KeyboardButton creation, making it easier to create keyboards.


noplanman:
I've just finished an initial version of the keyboard rewrites. Apart from supporting the keyboard object creation as it was till now, I've implemented a faster way of creating standard keyboards:
```php
/**
 * Gives 2 rows, each with 1 button.
 *
 * [           Button 1           ]
 * [           Button 2           ]
 */
$keyboard = new Keyboard('Button 1', 'Button 2');

/**
 * Gives 1 row, with 2 buttons.
 *
 * [   Button 1   ][   Button 2   ]
 */
$keyboard = new Keyboard(['Button 1', 'Button 2']);

/**
 * Gives 2 rows, with 1 and 2 buttons.
 *
 * [           Button 1           ]
 * [   Button 2   ][   Button 3   ]
 */
$keyboard = new Keyboard('Button 1', ['Button 2', 'Button 3']);

// or...
$keyboard = (new Keyboard('Button 1'))
    ->addRow('Button 2', 'Button 3');

// or with an empty keyboard initialisation...
$keyboard = (new Keyboard([]))
    ->addRow('Button 1')
    ->addRow('Button 2', 'Button 3');
```

Here a few other ways to do the same thing with added properties:
```php
$keyboard = new Keyboard(
    ['text' => 'Send Contact', 'request_contact' => true],
    ['text' => 'Send Location', 'request_location' => true]
);
$keyboard = new Keyboard(
    (new KeyboardButton('Send Contact'))->setRequestContact(true),
    (new KeyboardButton('Send Location'))->setRequestLocation(true)
);
```

Changing keyboard properties is easy:
```php
$keyboard
    ->setResizeKeyboard(true)
    ->setOneTimeKeyboard(true)
    ->setSelective(false);
```

`ForceReply` and `Hide` keyboards are easier now too:
```php
$keyboard = Keyboard::forceReply();
$keyboard = Keyboard::hide();
```
Note: These also return a `Keyboard` object, that can be manipulated the same way as any other keyboard.

Same system applies for `InlineKeyboard`.

This isn't 100% done yet and I need to refactor a bit and write the tests, but what do you think of this idea?

@akalongman @MBoretto @jacklul 
Please give me your feedback 👍

MBoretto:
Hello I would like to address this point:
all condition like: 
```$variable !== null```

Need  to be changed to:  

``` isset($variable) ``` 

see for example User.php entity
then we also need  to check carefully if all the updates are stored correctly in the database
You renamed the Unit dir in unit! 
Amazing work!

noplanman:
Conditions like `$variable !== null` make sense if we're checking for `null` only.
For other cases `isset($variable)` is ok, but should be used **with caution**, as `0, '', false, []` are all empty.

Wherever I check for `null` it normally has it's reason, but I will double check!

PR #314, opened 2016-10-08
implement getWebhookInfo method

theone74:
https://core.telegram.org/bots/api#getwebhookinfo


noplanman:
Hi @theone74 

Thanks for the PR! 🎉 

I will merge it into develop, but the code itself will change when I merge it into my pending PR (#291) which is about updating the whole Entity management.

PR #329, opened 2016-11-17
Use PSR-7 stream as file source

pavlyuts:
Makes possible to send all kind of files from memory, not from file on disk by usage of PSR-7 streams.

All we need is to create PSR-7 stream and pass it as a part of $data to any of sendXXXX method InputFile is allowed for by Bot API. The patch fixes exec method to handle request as multipart for both cases: file resource and stream.

The key should be set in accordance to the API call type: 'document', 'photo', 'audio', e.t.c.

PSR-7 metadata 'uri' must contain filename.

Code example:

    $stream = \GuzzleHttp\Psr7\stream_for( 'File contents here!', [ 'metadata' =>['uri'=>'filename.ext']]);

    $data = [];
    $data['chat_id'] = $chat_id;
    $data['caption'] = $caption;
    $data['document'] = $stream;
            
    Request::sendDocument($data);

The example will send as document the file named 'filename.ext' containing 'File contents here!'.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -181,7 +181,7 @@ class Request
         //Check for resources in data
         $contains_resource = false;
         foreach ($data as $item) {
-            if (is_resource($item)) {
+            if ((is_resource($item)) || ($item instanceof \GuzzleHttp\Psr7\Stream)) {
                 $contains_resource = true;
                 break;
             }


noplanman:
Cool, I'll check this out when I'm back home, as I'm travelling at the moment.

Could you also rebase your branch so that there's no conflict please, then I can merge.

pavlyuts:
HI!

Sorry, but i can't rebase the branch. I need this patch for my project, and I am not ready to have deal with every single update and mergw only 0.xx releases to my branch. 

Therefore I use branch started from 0.35 tag and the fix branch started there and alredy merged to my 'project used' branch. So, Git will not allow me to rebase.

I found the conflict reason, it comes from commit 5131894 and some previous changes. By the way, I am not agree that this commit is 'optimisation'. The old logic was to scan $data for resource and IF present to reformat $data multipart way. But after 'optimisation' you scan and format multipart in one place, and finally return source $data or multipart data depends of resource found or not. That means you do useless work because the most queries are NOT file send type and don't need multipart.

More, I guess the whole approach is wrong. We try to guess if we need use multipart query or no. But we exactly know it one step back. If we send file - we need multipart. If we send any other - we need simple query. So, I have no idea why don't just use two different methods like 'execute' and 'executeMultipart' and call each with respect of sending message type. Guess, we can send not file, but Telegram ID of the file, therefore we don't need multipart. 

However, Telegram API says that multipart is mandatory for file body upload. But multipart is also may bu used for ANY kind of call. So, it's simple just to use multipart only requests to Telegram and don't overcode the workframe.

Any way, you have to decline my pull request, it can't be merged now. To be compatible with your new vision of Send::execute I need to prepare the new one request for your new code.

PR #334, opened 2016-11-25
move hideKeyboard to removeKeyboard

MBoretto:
hide Keyboard is deprecated since Bot api 2.3
I tested it with my bot

### DIFF
--- a/src/Commands/AdminCommands/SendtochannelCommand.php
+++ b/src/Commands/AdminCommands/SendtochannelCommand.php
@@ -97,7 +97,7 @@ class SendtochannelCommand extends AdminCommand
                     $this->conversation->update();
 
                     $data['text']         = 'Insert the channel name: (@yourchannel)';
-                    $data['reply_markup'] = Keyboard::hide(['selective' => true]);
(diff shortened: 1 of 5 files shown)


noplanman:
From Telegram API docs:

> hide_keyboard will still work for a while for backward compatibility.

Maybe we can support both for now? Similar to [this](https://github.com/akalongman/php-telegram-bot/blob/3120727f2943e67ebc412ae90c6d733418c2cb65/src/Entities/Message.php#L88).

Also, it would make sense to add some kind of internal `deprecated` function to output a warning/error to the log, that an obsolete function is being used.

MBoretto:
I thought about it but is not the same case. For database if you don't have retrocompatibility you cannot replay the logs so you lose data. Instead this change has no impact on the data and there's no meaning to support it anymore in the library. Everybody need to change the method in the Commands if they use this.

noplanman:
I was thinking to give users time to change, not just suddenly stop supporting it.
Having said that, if they update to 0.36 they will need to change many other things anyway, so you're quite right.

I'll merge then, ok?

MBoretto:
ok!

PR #365, opened 2016-12-26
Botan improvements and misc fixes

jacklul:
Botan.io:
- moved to Guzzle
- user can set own timeout for Botan requests `$telegram->enableBotan('TOKEN', ['timeout' => 5])`, defaults to 3
- added shortener usage example
- stated potential usage for `public static $command` in docblock
- fixed botan not working with getUpdates

Other:
- Fixed `Call to a member function getBody() on null` happening on empty or no replies (currently visible on Windows, will be replaced with 'invalid response' exception which is better)
- Fixed code with PHP Code Beautifier and Fixer
- Added example usage of editMessageText 
(I think it would be a good idea to make more usage/command examples based on past closed issues)

### DIFF
--- a/examples/getUpdatesCLI.php
+++ b/examples/getUpdatesCLI.php
@@ -62,7 +62,9 @@ try {
     //$telegram->setUploadPath('../Upload');
 
     // Botan.io integration
+    // Second argument is optional maximum timeout
     //$telegram->enableBotan('your_token');
+    //$telegram->enableBotan('your_token', 3);
 
     // Handle telegram getUpdates request
     $serverResponse = $telegram->handleGetUpdates();
--- a/examples/getUpdatesCLI.php
+++ b/examples/getUpdatesCLI.php
@@ -64,7 +64,7 @@ try {
     // Botan.io integration
     // Second argument is optional maximum timeout
     //$telegram->enableBotan('your_token');
-    //$telegram->enableBotan('your_token', 3);
+    //$telegram->enableBotan('your_token', ['timeout' => 3]);
 
     // Handle telegram getUpdates request
     $serverResponse = $telegram->handleGetUpdates();
--- a/examples/getUpdatesCLI.php
+++ b/examples/getUpdatesCLI.php
@@ -62,7 +62,7 @@ try {
     //$telegram->setUploadPath('../Upload');
 
     // Botan.io integration
-    // Second argument is optional maximum timeout
+    // Second argument are options
     //$telegram->enableBotan('your_token');
     //$telegram->enableBotan('your_token', ['timeout' => 3]);
 
--- a/src/Botan.php
+++ b/src/Botan.php
@@ -194,7 +194,12 @@ class Botan
             $responseData = json_decode($response, true);
 
             if ($responseData['status'] !== 'accepted') {
-                TelegramLog::debug("Botan.io API replied with error:\n$response\n\n");
+                if (!empty($response)) {
+                    TelegramLog::debug("Botan.io track post failed, API reply:\n$response\n\n");
+                } else {
+                    TelegramLog::debug("Botan.io track post failed, API returned empty response!\n\n");
+                }
+
                 return false;
             }
 
@@ -245,7 +250,12 @@ class Botan
                 BotanDB::insertShortUrl($user_id, $url, $response);
                 return $response;
             } else {
-                TelegramLog::error("Botan.io URL shortening failed - API replied with error:\n$response\n\n");
+                if (!empty($response)) {
+                    TelegramLog::debug("Botan.io URL shortening failed for '$url', API reply:\n$response\n\n");
+                } else {
+                    TelegramLog::debug("Botan.io URL shortening failed for '$url', API returned empty response!\n\n");
+                }
+
                 return $url;
             }
         }
--- a/src/Botan.php
+++ b/src/Botan.php
@@ -194,11 +194,7 @@ class Botan
             $responseData = json_decode($result, true);
 
             if ($responseData['status'] !== 'accepted') {
-                if (!empty($result)) {
-                    TelegramLog::debug("Botan.io stats report failed: $result\n\n");
-                } else {
-                    TelegramLog::debug("Botan.io stats report failed: empty response!\n\n");
-                }
+                TelegramLog::debug('Botan.io stats report failed: ' . ($result ?: 'empty response') . "\n\n");
 
                 return false;
             }
@@ -247,11 +243,7 @@ class Botan
                 BotanDB::insertShortUrl($user_id, $url, $result);
                 return $result;
             } else {
-                if (!empty($result)) {
-                    TelegramLog::debug("Botan.io URL shortening failed for '$url': $result\n\n");
-                } else {
-                    TelegramLog::debug("Botan.io URL shortening failed for '$url': empty response!\n\n");
-                }
+                TelegramLog::debug('Botan.io URL shortening failed for \'' . $url . '\': ' . ($result ?: 'empty response') . "\n\n");
 
                 return $url;
             }
--- a/src/Botan.php
+++ b/src/Botan.php
@@ -166,11 +166,7 @@ class Botan
         }
 
         // In case there is no from field assign id = 0
-        if (isset($data['from']['id'])) {
-            $uid = $data['from']['id'];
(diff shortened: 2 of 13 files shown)


src/Commands/AdminCommands/SendtochannelCommand.php
```diff
@@ -81,7 +81,7 @@ public function execute()
         $notes = &$this->conversation->notes;
         !is_array($notes) && $notes = [];
 
-        $channels = (array)$this->getConfig('your_channel');
+        $channels = (array) $this->getConfig('your_channel');
```
noplanman: We should decide on a specific format and keep it that way, otherwise these little things will be overwritten all the time by each other.
At the moment there is some mix of this, some places are one way, others another.
I'll set up an issue for us to discuss this briefly, as it's not such a big deal...

src/Botan.php
```diff
@@ -181,33 +227,37 @@ public static function shortenUrl($url, $user_id)
...
+                )
+            );
 
-        $context  = stream_context_create($options);
-        $response = @file_get_contents($request, false, $context);
+            $response = (string) $response->getBody();
+        } catch (RequestException $e) {
+            $response = ($e->getResponse()) ? (string) $e->getResponse()->getBody() : '';
+        } finally {
+            if (!filter_var($response, FILTER_VALIDATE_URL) === false) {
```
noplanman: Just using `filter_var($response, FILTER_VALIDATE_URL) !== false` is easier to read.

src/Botan.php
```diff
@@ -181,33 +227,37 @@ public static function shortenUrl($url, $user_id)
...
-        if (!empty($cached[0]['short_url'])) {
-            return $cached[0]['short_url'];
-        }
 
-        $request = str_replace(
-            ['#TOKEN', '#UID', '#URL'],
-            [self::$token, $user_id, urlencode($url)],
-            self::$shortener_url
-        );
+        if (!empty($cached)) {
```
noplanman: `!empty()` is redundant here, as `$cached` gets defined above.
You could also just do:
```php
if ($cached = BotanDB::selectShortUrl($user_id, $url)) {
    return $cached;
}
```

src/Botan.php
```diff
@@ -20,82 +23,107 @@
...
         self::$command = '';
 
-        $obj  = json_decode($input, true);
+        // For now, this is the only way
+        $update_data = (array) $update;
+
         $data = [];
-        if (isset($obj['message'])) {
-            $data       = $obj['message'];
+        if ($update->getMessage()) {
```
noplanman: Would be nice if we could find a way to shorten this epic `if else` block and make it more dynamic.
Not priority though, more a though since you've touched code here 😃

src/Telegram.php
```diff
@@ -815,14 +816,15 @@ protected function ucfirstUnicode($str, $encoding = 'UTF-8')
...
      *
-     * @param  $token
+     * @param  string $token
+     * @param  integer $timeout
      *
      * @return \Longman\TelegramBot\Telegram
      * @throws \Longman\TelegramBot\Exception\TelegramException
      */
-    public function enableBotan($token)
+    public function enableBotan($token, $timeout = 3)
```
noplanman: I'm just wondering if we might want to have an array here for extra parameters, to prevent issues in the future if new parameters might become available.
Having an array, we just add indexes and previous code still works. Adding extra parameters to the method becomes difficult to manage.

Here an example of the problem I'm trying to explain:
https://github.com/akalongman/php-telegram-bot/pull/346/files#diff-04c6e90faac2675aa89e2176d2eec7d8L245

jacklul:
Which line?

src/Botan.php
```diff
@@ -20,152 +23,195 @@
...
+            'message' => 'Message',
+            'edited_message' => 'Edited Message',
+            'channel_post' => 'Channel Post',
+            'edited_channel_post' => 'Edited Channel Post',
+            'inline_query' => 'Inline Query',
+            'chosen_inline_result' => 'Chosen Inline Result',
+            'callback_query' => 'Callback Query'
+        ];
+
+        if (in_array($update_type, ['message', 'edited_message', 'channel_post', 'edited_channel_post', 'inline_query', 'chosen_inline_result', 'callback_query'], true)) {
```
noplanman: You could use this instead:
```php
if (array_key_exists($update_type, $update_object_names)) {...
```

Right?

jacklul: Oh, that's what you meant!

examples/getUpdatesCLI.php
```diff
@@ -62,7 +62,9 @@
     //$telegram->setUploadPath('../Upload');
 
     // Botan.io integration
+    // Second argument is optional maximum timeout
```
noplanman: Nitty-gritty: `Second argument are options`

*(same for hook.php)*

src/Botan.php
```diff
@@ -20,152 +23,195 @@
...
+                    'json' => $data
+                ]
+            );
 
-        if ($responseData['status'] !== 'accepted') {
-            TelegramLog::debug('Botan.io API replied with error: ' . $response);
-        }
+            $response = (string) $response->getBody();
+        } catch (RequestException $e) {
+            $response = ($e->getResponse()) ? (string) $e->getResponse()->getBody() : '';
```
noplanman: Looking at the Guzzle `RequestException` class, it seems the response always gets set.

Have you run into a situation where this wasn't the case?

Also, I think variable-name-wise, it would be better not to overwrite `$response` with the response body, as that would actually be `$response_body`.
Maybe call and check the `->getBody()` in the `finally` clause? Also to remove the duplication above.

*(same applies to the other similar code further down)*

jacklul: Yes, currently on Windows there is `can't call getBody() on null` error here and in `Request::execute()`.

src/Botan.php
```diff
@@ -187,17 +187,17 @@ public static function track(Update $update, $command = '')
...
-            $response = ($e->getResponse()) ? (string) $e->getResponse()->getBody() : '';
+            $result = $e->getMessage();
         } finally {
-            $responseData = json_decode($response, true);
+            $responseData = json_decode($result, true);
 
             if ($responseData['status'] !== 'accepted') {
-                if (!empty($response)) {
-                    TelegramLog::debug("Botan.io stats report failed, API reply: $response");
+                if (!empty($result)) {
```
noplanman: Well, if you're doing it like this, might as well shorten it all the way 😃
```php
TelegramLog::debug('Botan.io stats report failed: ' . ($result ?: 'empty response') . "\n\n");
```

*(same for the bottom one)*

src/Botan.php
noplanman: Oh right, ok... The way you've changed it in the last commit works for Windows nicely now too then?

jacklul: At least not throwing that error and stopping execution like it did before...

src/BotanDB.php
```diff
@@ -32,8 +32,8 @@ public static function initializeBotanDb()
     /**
```
noplanman: I think it would make sense to use the same parameter order like for `shortenUrl($url, $user_id)`:
so `selectShortUrl($url, $user_id)`

src/Botan.php
```diff
@@ -20,152 +23,191 @@
...
-        $options = [
-            'http' => [
-                'header'        => 'Content-Type: application/json',
-                'method'        => 'POST',
-                'content'       => json_encode($data),
-                'ignore_errors' => true,
-            ],
-        ];
+        // In case there is no from field assign id = 0
+        if (isset($data['from']['id'])) {
```
noplanman: Maybe shorten this to 1 line too? Up to you.

src/Botan.php
```diff
@@ -20,152 +23,191 @@
...
+            $uid = $data['from']['id'];
+        } else {
+            $uid = 0;
+        }
 
-        $context      = stream_context_create($options);
-        $response     = @file_get_contents($request, false, $context);
-        $responseData = json_decode($response, true);
+        try {
+            $response = self::$client->post(
```
noplanman: This whole block looks really nice and neat now, well done :1st_place_medal:

src/BotanDB.php
jacklul: That was on my mind but it seems to forgot about it, will do!

PR #368, opened 2016-12-29
Scrutinizer fixes

noplanman:
Fix most of the **Major** scrutinizer issues.

### DIFF
--- a/src/TelegramLog.php
+++ b/src/TelegramLog.php
@@ -10,10 +10,10 @@
 
 namespace Longman\TelegramBot;
 
-use Monolog\Logger;
-use Monolog\Handler\StreamHandler;
-use Monolog\Formatter\LineFormatter;
 use Longman\TelegramBot\Exception\TelegramLogException;
+use Monolog\Formatter\LineFormatter;
+use Monolog\Handler\StreamHandler;
+use Monolog\Logger;
 
 class TelegramLog
 {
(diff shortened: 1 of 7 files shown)


src/TelegramLog.php
```diff
@@ -147,11 +147,10 @@ public static function initDebugLog($path)
     public static function getDebugLogTempStream()
     {
         if (self::$debug_log_temp_stream_handle === null) {
-            if (self::isDebugLogActive()) {
-                self::$debug_log_temp_stream_handle = fopen('php://temp', 'w+');
-            } else {
+            if (!self::isDebugLogActive()) {
```
chuv1: Since it's just a return, is there a need of "{}"?
It could be wrote on one line, no?

I bag your pardon for this message, but I've read a lot of your code last days, and I really like it, it's really nice, clean and i'd even say beautiful, so I'm just asking. )

noplanman: @chuv1 Thanks for the nice complement! 😇 

Yes, you are correct, they could be removed. But as part of the style guide, we keep them even for `if` loops with just 1 line.

Let me explain why in a short example, using this exact code.
Say we have the current code as is, that returns `false` if the debug log isn't active, without the `{}` this time:
```php
if (!self::isDebugLogActive())
    return false;
```

Let's assume that we need to add some important function that needs to be called before returning `false`. This can quickly lead to an ugly bug:
```php
if (!self::isDebugLogActive())
    somethingImportant();
    return false;
```
In this case, only `somethingImportant()` is called inside the `if` clause, and `false` will **always** be returned by this method! 😱 

So it's better to just put the `{}` and be totally safe 🎉 

But of course you can do as you like, as long as you're aware of it and take special care, it should be fine.

PR #379, opened 2017-01-01
Fix for channel_post and edit_channel_post commands

jacklul:
_Seems that I've made small mistake, sorry!_


noplanman:
No problem, thanks for fixing it 😇

Issue #156, opened 2016-04-19
Task scheduler


jacklul:
For example if you want to send notifications to users at specified time and date.
This can be of course done currently by sending fake update but it would be nice to have a build-in task scheduler.

ruslanmedia:
You need add to cron

aleksraiden:
At my project, I'am using Disque (project from Redis author) and one simple php daemon to send messages.

jacklul:
Currently in testing: https://github.com/jacklul/php-telegram-bot/tree/crontab

PR #382, opened 2017-01-04
Crontab-friendly script

jacklul:
So, basically this will let the developers run their commands from crontab easily.

This essentialy is a solution for https://github.com/akalongman/php-telegram-bot/issues/156

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -830,4 +830,68 @@ class Telegram
 
         return $this;
     }
+
+    /**
+     * Run provided commands
+     *
+     * @param array $commands
+     * @param array $update
+     *
+     * @throws TelegramException
+     */
+    public function runCommands($commands, array $update = [])
+    {
+        if (!isset($commands) || !is_array($commands)) {
+            throw new TelegramException('No command(s) provided!');
+        }
+
+        $this->botan_enabled = false;   // Force disable Botan.io integration, we don't want to track self-executed commands!
+
+        $result = Request::getMe()->getResult();
+
+        if (!$result) {
+            throw new TelegramException('Received invalid getMe result!');
+        }
+
+        $bot_id   = $result->getId();
+        $bot_name = $result->getFirstName();
+
+        $update_template = [
+            'update_id' => 0,
+            'message' => [
+                'message_id' => 0,
+                'from' => [
+                    'id' => $bot_id,
+                    'first_name' => $bot_name
+                ],
+                'date' => time(),
+                'chat' => [
+                    'id' => $bot_id,
+                    'type' => 'private',
+                ],
+                'text' => ''
+            ]
+        ];
+
+        $update = array_merge($update_template, $update);
+
+        $this->enableAdmin($bot_id);    // Give bot access to admin commands
+        $this->getCommandsList();       // Load full commands list
+
+        foreach ($commands as $command => $parameter) {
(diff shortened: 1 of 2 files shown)


src/Telegram.php
```diff
@@ -830,4 +830,68 @@ public function enableBotan($token, array $options = [])
...
+
+        $bot_id   = $result->getId();
+        $bot_name = $result->getFirstName();
+
+        $update_template = [
+            'update_id' => 0,
+            'message' => [
+                'message_id' => 0,
+                'from' => [
+                    'id' => $bot_id,
```
noplanman: Prettify alignment 😇

src/Telegram.php
```diff
@@ -830,4 +830,68 @@ public function enableBotan($token, array $options = [])
...
+            ]
+        ];
+
+        $update = array_merge($update_template, $update);
+
+        $this->enableAdmin($bot_id);    // Give bot access to admin commands
+        $this->getCommandsList();       // Load full commands list
+
+        foreach ($commands as $command => $parameter) {
+            if (is_numeric($command)) {     // if array/key is not associative $command will be integer and $parameter will be the actual command!
```
noplanman: Just so I understand this correctly:
```php
$commands = [
    'whoami',
    'echo' => 'hello world!',
];
```
Would call `/whoami` and `/echo hello world!`, right?

If yes, I really like your thinking behind this, although this does limit to 1 single call of any single command with params, as the key is unique.
```php
$commands = [
    'echo' => 'will I be echoed?',
    'echo' => 'nope, just me!',
];
```

Might be an idea to just add the whole command as array elements, like so:
```php
$commands = [
    '/echo will I be echoed?',
    '/echo yes! and so will I :-)',    
];
```
This way, also "normal" text commands could be sent theoretically, although the way it's coded now this won't work.

jacklul:
But I always agree with them, they always are an improvement I just didn't came up with!

noplanman:
Finally gotten around to test this... Nice work!

I see that it doesn't add anything to the DB, obviously, as it gets an update ID of `0`.
Was just wondering if there might be a situation where one would want that. Not sure about implementation then to keep it nice and separate from the main updates.

Also, it would be great if you could make a short wiki entry explaining what this does.
Namely, that this cron can be used to have the bot send messages to itself.

Regarding the short command example you give at the top of the `cron.php` file, it would be great to put more meaningful ones, as sending `/whoami` or `/echo vegetable` doesn't really *do* anything.
I assume you use this to call maintenance related commands?

Apart from that I'm happy to merge this, will wait for your response 👌

jacklul:
> (...)
> Was just wondering if there might be a situation where one would want that. Not sure about implementation then to keep it nice and separate from the main updates.

I have no idea why would someone want to put 'fake' updates coming from crontab into DB, at first I've been using setCustomInput and sent fake update to run my scheduled commands, but it was literally flooding db with useless things (running 3 commands every minute),

> Also, it would be great if you could make a short wiki entry explaining what this does.

Will do!

> Regarding the short command example you give at the top of the cron.php file, it would be great to put more meaningful ones, as sending /whoami or /echo vegetable doesn't really do anything.
I assume you use this to call maintenance related commands?

Well, I didn't want to put non-existing commands there as an example...

noplanman:
Awesome wiki entry, very nice!

PR #387, opened 2017-01-08
Debug command

noplanman:
Add a simple debug command that outputs general info about the bot and the server environment.

@jacklul Let me know if there's anything else you'd want in here.
Was thinking of some stats from the DB, but not sure how useful that is for debugging.

### DIFF
--- a/src/Commands/AdminCommands/DebugCommand.php
+++ b/src/Commands/AdminCommands/DebugCommand.php
@@ -1,0 +1,93 @@
+<?php
+/**
+ * This file is part of the TelegramBot package.
+ *
+ * (c) Avtandil Kikabidze aka LONGMAN <akalongman@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Longman\TelegramBot\Commands\AdminCommands;
+
+use Longman\TelegramBot\Commands\AdminCommand;
+use Longman\TelegramBot\DB;
+use Longman\TelegramBot\Request;
+
+/**
+ * Admin "/debug" command
+ */
+class DebugCommand extends AdminCommand
+{
+    /**
+     * @var string
+     */
+    protected $name = 'debug';
+
+    /**
+     * @var string
+     */
+    protected $description = 'Debug command to help find issues';
+
+    /**
+     * @var string
+     */
+    protected $usage = '/debug';
+
+    /**
+     * @var string
+     */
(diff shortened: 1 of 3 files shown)


src/Commands/AdminCommands/DebugCommand.php
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        $debug_info[] = '_Webhook Info_:';
+        try {
+            $webhook_info = json_encode(json_decode(Request::getWebhookInfo(), true)['result'], JSON_PRETTY_PRINT);
+        } catch (\Exception $e) {
+            $webhook_info = sprintf('Failed to get webhook info! (%s)', $e->getMessage());
+        }
+        $debug_info[] = sprintf("```\n%s\n```", $webhook_info);
+
```
jacklul: For getUpdates users showing webhook info will be useless, I believe you can make a check on that with `Request::getInput()`.

chuv1:
I got "Method getPdo not found" error...

noplanman:
@chuv1 You'll need to checkout the develop branch, as it has been updated with methods that this command relies on. You can't just copy paste this command.

chuv1:
I can. ) 
I just commented out that PDO stuff. )

chuv1:
Btw, "pending_update_count": in webhookInfo info returns 1, but getting getWebhookInfo through api.telegram.org with browser afterward shows 0. 
It's this /debug update is pending, right?

noplanman:
> It's this /debug update is pending, right?

Yes, that's it. Feels very hacky to just do `-1` there 😕
But that might be the best, to not confuse users...

noplanman:
@jacklul 

- [ ] I would make sure this command does not return any sensitive data in groups, I think it should only work in private chats!

> I think the bold option with extra space after looks pretty!
- [x] I played around with different ways of displaying this, keeping the focus on the important parts, the values. But I must say that your suggestion does look pretty sweet 😃 **Done!**
Btw. what client are you using?! So weird that the second line is bold, check the code 

- [x] Download/Upload paths **Done!**
- [ ] A way to list all command paths would also be nice (in cases when user's commands doesn't get loaded and he has no idea why)

(Thanks for the input! Will be updating this comment!)

jacklul:
> Btw. what client are you using?! So weird that the second line is bold, check the code

Screenshot contained already modified output to visualize what I wrote later.  :simple_smile:

noplanman:
@jacklul Would appreciate another look, thanks!

Issue #157, opened 2016-04-22
Make bot follow the limits


jacklul:
sendToActiveChats() can easily block your bot for few minutes if you go over the limit

https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this

aliemam:
Tanx...

MBoretto:
What I'm thinking that we can create 'something' (maybe related to the task manager https://github.com/akalongman/php-telegram-bot/issues/156) that sends a little amount of messages every time an update is received.
Then with the command /sendtochats you can set a message to send and see the overall process 150/2500 user reached.
What do you think?

commfort:
@jacklul @MBoretto not sure if limits  should be part of API wrapper. You need to leave it for custom logic. 
Let's say my bot has billions of users in groups and privates. Than I need to give priority for group admins over plain visitors. And I need to think about action limitation for single user and even to have ability to ban/unban spam users. 

On other hand I can prefer implement limits/task manager using memcache or rabbit, or by storing history in mysql and scheduling some cron job.

jacklul:
Currently testing: https://github.com/jacklul/php-telegram-bot/tree/limiter_test

PR #397, opened 2017-01-21
Request limiter - prevent hitting Telegram's API limits

jacklul:
#157

So, I've been using this implementation for a while now and I don't see any issues so far, neither any exceptions about Telegram returning response that I've hit the limit.  

By default disabled because it can be heavy on the database. 
Will delay requests using `sleep(1);`.

As mentioned in https://core.telegram.org/bots/faq#my-bot-is-hitting-limits-how-do-i-avoid-this
- No more than 1 message / second (chat)
- No more than 30 messages / second (globally)
- No more than 20 messages / minute (groups and channels)

This also applies to message editing! Does not apply to methods like getChat or sendChatAction.

Currently I also included 'retry' mode, sometimes if the bot is heavily used 60 seconds wait might not be enough... should I remove this and simply give it only 60 seconds, no more? Or leave it as is?

**Will appreciate anyone to take a look at this!
Suggestions are welcome!**

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -1029,21 +1029,30 @@ class Request
                 'editMessageReplyMarkup',
             ];
 
-            if ((isset($data['chat_id']) || isset($data['inline_message_id'])) && in_array($action, $limited_methods)) {
+            $chat_id = isset($data['chat_id']) ? $data['chat_id'] : null;
+            $inline_message_id = isset($data['inline_message_id']) ? $data['inline_message_id'] : null;
+
+            if (($chat_id || $inline_message_id) && in_array($action, $limited_methods)) {
                 $timeout = 60;
 
                 while (true) {
                     if ($timeout <= 0) {
-                        throw new TelegramException('Timed out while waiting for a request slot!');
+                        throw new TelegramException('Timed out while waiting for a request spot!');
                     }
 
-                    $requests = DB::getTelegramRequestCount((isset($data['chat_id']) ? $data['chat_id'] : null), (isset($data['inline_message_id']) ? $data['inline_message_id'] : null));
+                    $requests = DB::getTelegramRequestCount($chat_id, $inline_message_id);
 
-                    if ($requests['LIMIT_PER_SEC'] <= 0 && ((isset($data['inline_message_id']) && $requests['LIMIT_PER_SEC_ALL'] < 30) || (isset($data['chat_id']) && $data['chat_id'] > 0 && $requests['LIMIT_PER_SEC_ALL'] < 30) || (isset($data['chat_id']) && $data['chat_id'] < 0 && $requests['LIMIT_PER_MINUTE'] < 20))) {
+                    if (
+                        $requests['LIMIT_PER_SEC'] == 0 &&  // No more than one message per second inside a particular chat
+                        (
+                            (($chat_id > 0 || $inline_message_id) && $requests['LIMIT_PER_SEC_ALL'] < 30) ||  // No more than 30 messages per second globally
+                            ($chat_id < 0 && $requests['LIMIT_PER_MINUTE'] < 20)  // No more than 20 messages per minute for group chats
+                        )
+                    ) {
                         break;
                     }
 
-                    $timeout = $timeout - 1;
+                    $timeout--;
                     sleep(1);
                 }
 
--- a/src/Request.php
+++ b/src/Request.php
@@ -1027,6 +1027,7 @@ class Request
 
             if ((isset($data['chat_id']) || isset($data['inline_message_id'])) && in_array($action, $limited_methods)) {
                 $timeout = 60;
+                $tick = 500000; //msec
 
                 if (!is_numeric($data['chat_id'])) {
                     $data['chat_id'] = 0;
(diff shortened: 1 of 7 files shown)


src/Request.php
```diff
@@ -976,4 +985,70 @@ public static function getWebhookInfo()
...
+                'editMessageText',
+                'editMessageCaption',
+                'editMessageReplyMarkup',
+            ];
+
+            if ((isset($data['chat_id']) || isset($data['inline_message_id'])) && in_array($action, $limited_methods)) {
+                $timeout = 60;
+
+                while (true) {
+                    if ($timeout <= 0) {
```
noplanman: You could also do `$timeout--` here instead of `$timeout = $timeout - 1;` further down.

nesttle:
Looks awesome, IMHO this should be implemented by default on next releases.

src/Request.php
```diff
@@ -976,4 +985,70 @@ public static function getWebhookInfo()
...
+                        throw new TelegramException('Timed out while waiting for a request slot!');
+                    }
+
+                    $requests = DB::getTelegramRequestCount((isset($data['chat_id']) ? $data['chat_id'] : null), (isset($data['inline_message_id']) ? $data['inline_message_id'] : null));
+
+                    if ($requests['LIMIT_PER_SEC'] <= 0 && ((isset($data['inline_message_id']) && $requests['LIMIT_PER_SEC_ALL'] < 30) || (isset($data['chat_id']) && $data['chat_id'] > 0 && $requests['LIMIT_PER_SEC_ALL'] < 30) || (isset($data['chat_id']) && $data['chat_id'] < 0 && $requests['LIMIT_PER_MINUTE'] < 20))) {
```
jacklul: I could transform it into multi-line statement with comments, should be fine?

noplanman: For this it would make sense to add variables for `$data['chat_id']` and `$data['inline_message_id']` as they're also used further up.

So instead of all the `isset() && ...` you could just use the variables directly.

Will think about this exact spot and how to make it more readable 👍

jacklul:
```php
$chat_id = isset($data['chat_id']) ? $data['chat_id'] : null;
$inline_message_id = isset($data['inline_message_id']) ? $data['inline_message_id'] : null;

if (($chat_id || $inline_message_id) && in_array($action, $limited_methods)) {
	$timeout = 60;

	while (true) {
		if ($timeout <= 0) {
			throw new TelegramException('Timed out while waiting for a request spot!');
		}

		$requests = DB::getTelegramRequestCount($chat_id, $inline_message_id);

		if (
			$requests['LIMIT_PER_SEC'] == 0 &&  // No more than one message per second inside a particular chat
			(
				(($chat_id > 0 || $inline_message_id) && $requests['LIMIT_PER_SEC_ALL'] < 30) ||  // No more than 30 messages per second globally
				($chat_id < 0 && $requests['LIMIT_PER_MINUTE'] < 20)  // No more than 20 messages per minute allowed for group chats
			)
		) {
			break;
		}

		$timeout--;
		sleep(1);
	}

	DB::insertTelegramRequest($action, $data);
}
```

Seems to look a bit better, what you think?

noplanman: Yep, looks a lot better, nice 👌

src/Request.php
```diff
@@ -1029,21 +1029,30 @@ private static function limitTelegramRequests($action, array $data = [])
                 'editMessageReplyMarkup',
             ];
 
-            if ((isset($data['chat_id']) || isset($data['inline_message_id'])) && in_array($action, $limited_methods)) {
```
nesttle: I love one-line code instead multiline

noplanman: @nesttle Me too! (as long as it's readable :wink:)

src/Request.php
```diff
@@ -976,4 +985,76 @@ public static function getWebhookInfo()
...
+                $timeout = 60;
+
+                while (true) {
+                    if ($timeout <= 0) {
+                        throw new TelegramException('Timed out while waiting for a request spot!');
+                    }
+
+                    $requests = DB::getTelegramRequestCount($chat_id, $inline_message_id);
+
+                    if ($requests['LIMIT_PER_SEC'] == 0  // No more than one message per second inside a particular chat
```
noplanman: Any way I can (realistically) test these limits?

src/Request.php
```diff
@@ -976,4 +985,76 @@ public static function getWebhookInfo()
...
+                        || ($chat_id < 0 && $requests['LIMIT_PER_MINUTE'] < 20))
+                    ) {
+                        break;
+                    }
+
+                    $timeout--;
+                    sleep(1);
+                }
+
+                DB::insertTelegramRequest($action, $data);
```
noplanman: I noticed that entries once added, never get deleted from the DB once handled.
Shouldn't they be cleared again after being processed?

noplanman:
Busy trying to test this and get it merged. Hope to get it in soon 😊

src/Request.php
jacklul: But if I remove them next requests will think they are ok to go. 
I could add a query to remove entries older than 1 minute but I don't think it will be good practice to run such query every second (or more under heavy load)...?

src/Request.php
jacklul: Doing it alone will be quite hard, you can test the one msg per sec and no more than 20 per minute in groups rules easily, 30 per sec globally will be pretty hard to achieve alone. 

Spamming command with callback button editing own message should be good testing ground.

src/Request.php
noplanman: Hmm, I'm just thinking long term here. As this is a temporary table, it just gets bigger and bigger, in some cases pretty quick, right?
So some sort of cleanup would be good I think.

Maybe something like:
```php
if (mt_rand(0, 999) === 42) {
    do_cleanup();
}
```

Or maybe I'm going too far here and this isn't really a problem...

jacklul: I guess, in my case I'm cleaning the table every 5 minutes, set as a task in Mysql. 

I have no 'good' idea how to implement cleanup inside this code without adding potential delays or performance losses.

noplanman: Ok, then let's leave this as is for the moment and passively think of something :wink:

noplanman:
Didn't find the time to thoroughly test this, but I think it should be ok to merge.
If no one has any additional comments, I'll go ahead and merge soon.

PR #405, opened 2017-02-07
Telegram Emoji Icons added

dzarezenko:
Supported by Telegram Emoji Icons unicode values added for fast usage.

### DIFF
--- a/src/Tools/Icons.php
+++ b/src/Tools/Icons.php
@@ -1,0 +1,1760 @@
+<?php
+/**
+ * This file is part of the TelegramBot package.
+ *
+ * (c) Avtandil Kikabidze aka LONGMAN <akalongman@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Longman\TelegramBot\Tools;
+
+/**
+ * Supported by Telegram Emoji Icons unicode values.
+ */
+class Icons {
+
+    const AERIAL_TRAMWAY = "\xf0\x9f\x9a\xa1";
+
+    const AIRPLANE = "\xe2\x9c\x88";
+
+    const AIRPLANE_DEPARTURE = "\xf0\x9f\x9b\xab";
+
+    const AIRPLANE_ARRIVING = "\xf0\x9f\x9b\xac";
+
+    const ALARM_CLOCK = "\xe2\x8f\xb0";
+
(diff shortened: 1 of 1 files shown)


noplanman:
I've been thinking of this and maybe just using the [spatie/emoji](https://packagist.org/packages/spatie/emoji) library for that, to keep it separate.

I came to the conclusion that this shouldn't be part of this library at all though, since this will only be needed for custom commands. In which case users should just use it in their personal projects by adding it to their `composer.json`.

@jacklul @MBoretto @akalongman 
What's your view on this?

akalongman:
@noplanman Telegram emojies is supported by https://packagist.org/packages/spatie/emoji ?

noplanman:
Does Telegram require special handling?

I recall just using the `Emoji::grinningFace();` etc. for Emojis in PhpStorm (which doesn't support native Emojis yet).

akalongman:
@noplanman can you please test if spatie/emoji emojies works with telegram?

dzarezenko:
Sorry to interrupt but if we will have this class inside this package, then we can use it's constants in the `Longman\TelegramBot\Commands\UserCommands\WeatherCommand` for example instead of hard-coded values (`getWeatherString()` method). 

And it will be possible without external dependencies.

noplanman:
@akalongman Yes, works no problem.

@dzarezenko That's true. I guess the question would then arise if we actually want any default commands as part of the library or not 😕

akalongman:
@dzarezenko thank you for your advice, maybe you are right

@noplanman feel free to decide :)

noplanman:
Honestly, I think default commands should be moved out of the library into the `examples` folder. This will solve all the issues of having to make dummy commands to get rid of them.

@jacklul Would appreciate your input on this. Adding libraries/features that are only used very sparingly doesn't make sense in my view and just adds bloat.

jacklul:
I personally think there is no need for this, if someone wants to use this kind of way of handling emojis he can simply include it with composer for the project.

Take a look here: https://github.com/jacklul/php-telegram-bot/tree/no_default_commands for test I've moved all commands to examples and included them by default as command path, this however breaks unit tests I have no idea how to fix!

noplanman:
@jacklul Ok, so I'll close off here. Will check out your branch and build on it, then you can PR when it's more or less ready.

@dzarezenko If you need Emoji for your project, just use the spatie/emoji project. It's complete and takes care of everything for you 👍

PR #423, opened 2017-03-07
Fix #421 (Infinity loop after forwarding message from channel)

### LINKED ISSUE #421: Infinity loop after forwarding message from channel
### Required Information

- PHP version: PHP 5.6.30
- PHP Telegram Bot version: 0.40.0
- Using MySQL database: yes
- Update Method: getUpdates
- Self-signed certificate:
- RAW update (if available):

### Expected behaviour
1. Forwarding message from channel.
2. Bot processing message.
3. Bot checking next update with incremented offset

### Actual behaviour
1. Forwarding message from channel.
2. Bot processing message.
3. Bot checking next update with same offset.
4. Bot processing same message.
5. Bot checking next update with same offset.
...
infinity loop why bot will get any new message (not forwarded from channel!)

### Steps to reproduce
1. Run getUpdatesCLI.php
2. Create chat with you bot
3. Join any channel (not a group!)
4. Forward any message from channel to you bot
5. Check logs… you will see:
"2017-03-07 01:21:36 - Processed 1 updates
2017-03-07 01:21:36 - Processed 1 updates
2017-03-07 01:21:36 - Processed 1 updates
2017-03-07 01:21:36 - Processed 1 updates"

### Extra details
I have not enough time to find how to fix the problem but its looks like it something around DB::insertMessageRequest and DB::insertTelegramUpdate. It’s looks like telegram-bot not inserting info about forwarded message to the telegram_update table in the DB because violating constrain message_ibfk_4 (but all operation making with INSERT IGNORE and this why everything happining in «silence»)

SerafimPikalov:
Small fix of copy/paste mistake. Just set right class (is must be Chat, but it was User) to forward_from_chat field for class Message.

### DIFF
--- a/src/Entities/Message.php
+++ b/src/Entities/Message.php
@@ -55,7 +55,7 @@ class Message extends Entity
             'from'              => User::class,
             'chat'              => Chat::class,
             'forward_from'      => User::class,
-            'forward_from_chat' => User::class,
+            'forward_from_chat' => Chat::class,
             'reply_to_message'  => self::class,
             'entities'          => MessageEntity::class,
             'audio'             => Audio::class,


noplanman:
Great catch, thanks a million! 🎉

PR #430, opened 2017-03-12
add functionality to allow hiding commands in /help command

lichtscheu:
add `protected $showInHelp` var to `Command.php`

add `public function getShowInHelp()` to `Command.php`

add  `if($command->getShowInHelp()) {}` to wrap single command output at `HelpCommand.php`

with `protected $showInHelp = false;` insert into the command,it is not listed with /help command.

I find this very useful, cause i have some commands i only use over cronjob and don't want this commands to show up at the help command.

### DIFF
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -60,6 +60,13 @@ abstract class Command
     protected $usage = 'Command usage';
 
     /**
+     * Show in Help
+     *
+     * @var bool
+     */
+     protected $showInHelp = true;
+
+    /**
      * Version
      *
      * @var string
@@ -251,6 +258,16 @@ abstract class Command
     {
         return $this->name;
     }
+
+    /*
+     * Get Show in Help
+     *
+     * @return bool
+     */
+     public function getShowInHelp()
+     {
+     	return $this->showInHelp;
+     }
 
     /**
      * Check if command is enabled
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -64,7 +64,7 @@ abstract class Command
      *
      * @var bool
      */
-     protected $show_in_help = true;
+    protected $show_in_help = true;
 
     /**
      * Version
@@ -264,10 +264,10 @@ abstract class Command
      *
      * @return bool
      */
-     public function showInHelp()
-     {
+    public function showInHelp()
+    {
         return $this->show_in_help;
-     }
+    }
 
     /**
      * Check if command is enabled
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -259,7 +259,7 @@ abstract class Command
         return $this->name;
     }
 
-    /*
+    /**
      * Get Show in Help
      *
      * @return bool
--- a/src/Commands/Command.php
+++ b/src/Commands/Command.php
@@ -266,7 +266,7 @@ abstract class Command
      */
      public function showInHelp()
      {
-     	return $this->show_in_help;
+        return $this->show_in_help;
      }
 
     /**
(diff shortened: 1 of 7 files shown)


src/Commands/UserCommands/HelpCommand.php
```diff
@@ -69,11 +69,13 @@ public function execute()
             );
 
             foreach ($command_objs as $command) {
-                $text .= sprintf(
+                if($command->getShowInHelp()) {
```
noplanman: Instead of checking for `true` and putting all the code into the `if` block, it's nicer to test for `false` and just continue with the next element.
e.g.
```php
if (!$command->showInHelp()) {
    continue;
}
```
Also, remember to keep your code nice and clean, so make the indentation correct 👍

src/Commands/Command.php
```diff
@@ -60,6 +60,13 @@
     protected $usage = 'Command usage';
 
     /**
+     * Show in Help
+     *
+     * @var bool
+     */
+     protected $showInHelp = true;
```
noplanman: For properties, we prefer snake case, i.e. `$show_in_help`

👍 for default being `true`

src/Commands/Command.php
```diff
@@ -252,6 +259,16 @@ public function getName()
         return $this->name;
     }
 
+    /*
+     * Get Show in Help
+     *
+     * @return bool
+     */
+     public function getShowInHelp()
```
noplanman: Instead of `getShowInHelp`, just use `showInHelp`. Similar to `isEnabled`, it makes it much easier to "read" the code.

Compare:
```php
$command->getShowInHelp();
```
to
```php
$command->showInHelp();
```
See what I mean? 😃

lichtscheu:
Thanks for the awesome review =)

src/Commands/UserCommands/HelpCommand.php
```diff
@@ -69,11 +69,16 @@ public function execute()
             );
 
             foreach ($command_objs as $command) {
+                if($command->showInHelp()) {
```
noplanman: Not sure you tested this one, give it a go and see if this does what you want... 😉 
Also, there should be a space after the `if` like so: `if (...)`

src/Commands/UserCommands/HelpCommand.php
```diff
@@ -69,11 +69,16 @@ public function execute()
...
+                    continue;
+                }
+
                 $text .= sprintf(
-                    '/%s - %s' . PHP_EOL,
```
noplanman: The indentation was correct, so please revert this part.

src/Commands/Command.php
```diff
@@ -252,6 +259,16 @@ public function getName()
         return $this->name;
     }
 
+    /*
```
noplanman: This needs to be `/**` for docblock to pick it up correctly.

Also, check the number of spaces and compare it with the others.
e.g.:
```php
/**
 * There is an extra space before the *s on these lines!
 */
```

src/Commands/Command.php
```diff
@@ -252,6 +259,16 @@ public function getName()
...
+     * Get Show in Help
+     *
+     * @return bool
+     */
+     public function showInHelp()
+     {
+     	return $this->show_in_help;
```
noplanman: Indent by 4 spaces, always.
What editor do you use? Make sure it has an [editorconfig](http://editorconfig.org/) plugin installed.

noplanman:
@lichtscheu Also, notice that the build failed. When you click the `Details` button, it brings you to the Travis CI page, showing you what went wrong. It's an awesome help to make sure that everything is working properly.

Also, your changes will need some new tests, but I'll do those an PR to your branch so that you can see how they're done. If you feel like trying yourself, feel free to do so!

Issue #436, opened 2017-03-21
InlineKeyboard and handler - simple example


MyZik:
### Required Information

- PHP version: 5.6
- PHP Telegram Bot version: last
- Using MySQL database: yes
- Update Method: Webhook
- Self-signed certificate: yes

Hi everyone. How can I make very simple InlineKeyboard with text message update?
I have a message "Choose category" and under it there should be three buttons of InlineKeyboard: "Oranges", "Apples", "Pears". When I select a category, the message text must be updated. How can this be realized? Please show the code that corresponds to my example.
I delved into the files "InlineKeyboard" and "Inlinequery" but code is too complicated for understanding...
Help me pls!

noplanman:
Hi @MyZik 

Right, here is an example which you can start playing around with:

*CallbackqueryCommand.php*
```php
<?php

namespace Longman\TelegramBot\Commands\SystemCommands;

use Longman\TelegramBot\Commands\AdminCommands\CategoryCommand;
use Longman\TelegramBot\Commands\SystemCommand;
use Longman\TelegramBot\Request;

class CallbackqueryCommand extends SystemCommand
{
    protected $name = 'callbackquery';
    protected $description = 'Reply to callback query';
    protected $version = '1.0.0';

    public function execute()
    {
        $update         = $this->getUpdate();
        $callback_query = $update->getCallbackQuery();
        $callback_data  = $callback_query->getData();

        // Only do something for the 'category' selection.
        if (strpos($callback_data, 'category_') !== 0) {
            return Request::emptyResponse();
        }

        // Get the real category from the callback_data.
        $category = substr($callback_data, strlen('category_'));

        return Request::editMessageText([
            'chat_id'    => $callback_query->getMessage()->getChat()->getId(),
            'message_id' => $callback_query->getMessage()->getMessageId(),
            'text'       => CategoryCommand::$categories[$category],
        ]);
    }
}
```
*CategoryCommand.php*
```php
<?php

namespace Longman\TelegramBot\Commands\AdminCommands;

use Longman\TelegramBot\Commands\UserCommand;
use Longman\TelegramBot\Entities\InlineKeyboard;
use Longman\TelegramBot\Entities\InlineKeyboardButton;
use Longman\TelegramBot\Request;

class CategoryCommand extends UserCommand
{
    protected $name = 'category';
    protected $description = 'Choose a category, inline.';
    protected $usage = '/category';
    protected $version = '1.0.0';

[...]

PR #438, opened 2017-03-24
Inline keyboard pagination

noplanman:
This PR offers a simple way to add an inline pagination keyboard.

userlist

The idea has been around for a while and #436 gave me the kick to implement it.

**How it works**
Example command `/basicpages`
basicpages

`BasicpagesCommand.php`
In the command, implement the `InlineKeyboardPaginator`, populate the required static methods, set `reply_markup` to the paginator, done!
```php
<?php

namespace Longman\TelegramBot\Commands\UserCommands;

use Longman\TelegramBot\Commands\UserCommand;
use Longman\TelegramBot\Entities\InlineKeyboard;
use Longman\TelegramBot\Entities\InlineKeyboardPaginator;

class BasicpagesCommand extends UserCommand implements InlineKeyboardPaginator
{
    protected $name = 'basicpages';
    protected $description = 'Basic pages for inline keyboard pagination';
    protected $usage = '/basicpages';
    protected $version = '1.0.0';
    protected static $max_pages = 5; // Assigned dynamically in a real-life scenario!

    public static function getCallbackDataId()
    {
        return 'basicpages';
    }

    public static function getPagination($current_page)
    {
        return InlineKeyboard::getPagination(self::getCallbackDataId(), $current_page, self::$max_pages);
    }

    public static function getOutput($current_page)
    {
        return sprintf('Page %d of %d', $current_page, self::$max_pages);
    }

    public function execute()
    {
        // Using pagination
        $data = [
            'chat_id'      => $this->getMessage()->getChat()->getId(),
            'text'         => self::getOutput(1),
            'reply_markup' => self::getPagination(1),
        ];

        return Request::sendMessage($data);
    }
}
```

`CallbackqueryCommand.php`
In the callback query handler, we just need to get the current page number and can then edit the message to display the updated data.
```php
<?php

namespace Longman\TelegramBot\Commands\SystemCommands;

use Longman\TelegramBot\Commands\SystemCommand;
use Longman\TelegramBot\Commands\UserCommands\BasicpagesCommand;
use Longman\TelegramBot\Entities\InlineKeyboard;

class CallbackqueryCommand extends SystemCommand
{
    protected $name = 'callbackquery';
    protected $description = 'Reply to callback query';
    protected $version = '1.0.0';

    public function execute()
    {
        $callback_data = $this->getUpdate()->getCallbackQuery()->getData();

        if (strpos($callback_data, 'basicpages') === 0) {
            return $this->handleBasicPages();
        }

        return Request::emptyResponse();
    }

    protected function handleBasicPages()
    {
        $callback_query = $this->getUpdate()->getCallbackQuery();
        $callback_data  = $callback_query->getData();

        // Get the page from the callback_data.
        $current_page = InlineKeyboard::getPageFromCallbackData($callback_data);

        return Request::editMessageText([
            'chat_id'      => $callback_query->getMessage()->getChat()->getId(),
            'message_id'   => $callback_query->getMessage()->getMessageId(),
            'text'         => BasicpagesCommand::getOutput($current_page),
            'reply_markup' => BasicpagesCommand::getPagination($current_page),
        ]);
    }
}
```

**What would be nice to have changed to make this work even better**:
- Public static properties for commands, so that they can be accessed from everywhere, thus removing the requirement for `getCallbackDataId()`
- It would be brilliant to add some kind of `registerCallbackQueryHandler` method to the default system `CallbackqueryCommand` class. This could allow adding handlers without having to copy the command and adding it to a custom commands folder.

**Thoughts**:
The way it is now, when on the first or last page, only 3 buttons get displayed.
first-page
Would it make sense to always display 5 if there are enough pages?
Also, what if there are 100 pages, should there be a second row of buttons for greater page jumps?
big-keyboard

**Your opinion!**
Please give this a try and give me your feedback!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/) and this p
 
 ## [0.41.0] - 2017-03-25
 ### Added
-- `$showInHelp` attribute for commands, to set if it should be displayed in the `/help` command.
+- `$show_in_help` attribute for commands, to set if it should be displayed in the `/help` command.
 - Link to new Telegram group: `https://telegram.me/PHP_Telegram_Bot_Support`
 - Introduce change log.
 
--- a/src/Entities/InlineKeyboard.php
+++ b/src/Entities/InlineKeyboard.php
@@ -21,7 +21,7 @@ class InlineKeyboard extends Keyboard
      * Get an inline pagination keyboard.
      *
      * - $callback_data is an ID for the CallbackqueryCommand, to know where the request comes from.
-     * It must contain a '%d' placeholder for the page number. If no placeholder is found, the ID is automatically appended with '_page_%d'
+     * The ID is automatically appended with '_page_%d' to filter out the selected page number.
      *
      * - $labels allows for custom button labels, using '%d' placeholders.
      * Default:
@@ -46,9 +46,7 @@ class InlineKeyboard extends Keyboard
      */
     public static function getPagination($callback_data, $current_page, $max_pages, array $labels = [])
     {
-        if (strpos($callback_data, '%d') === false) {
-            $callback_data .= '_page_%d';
-        }
+        $callback_data .= '_page_%d';
 
         // Merge labels with defaults.
         $labels = array_merge([
@@ -98,5 +96,17 @@ class InlineKeyboard extends Keyboard
         }
 
         return new InlineKeyboard($buttons);
+    }
+
+    /**
+     * Extract the page number from the passed callback data.
+     *
+     * @param string $callback_data
+     *
+     * @return int
+     */
+    public static function getPageFromCallbackData($callback_data)
+    {
+        return (int) preg_replace('/.*_page_(\d+)$/', '$1', $callback_data);
     }
 }
--- a/src/Entities/InlineKeyboard.php
+++ b/src/Entities/InlineKeyboard.php
@@ -17,4 +17,86 @@ namespace Longman\TelegramBot\Entities;
  */
 class InlineKeyboard extends Keyboard
 {
+    /**
+     * Get an inline pagination keyboard.
+     *
+     * - $callback_data is an ID for the CallbackqueryCommand, to know where the request comes from.
+     * It must contain a '%d' placeholder for the page number. If no placeholder is found, the ID is automatically appended with '_page_%d'
+     *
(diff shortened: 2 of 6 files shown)


jacklul:
This is cool! 
Will probably make doing stuff like this easier for people!

**Some suggestions/ideas:**

- Maybe this could be modified to be a little more 'universal' so user could use alphabetical pagination?

- Would be nice if we could have few types of button layouts to choose from! Like even simple ones:

‹ | 2 |  ›
--- | --- | ---

« | ‹ | › | » 
--- | --- | --- | ---

In that case 'extended button layout':

> Also, what if there are 100 pages, should there be a second row of buttons for greater page jumps?

Definitely!

**Additional ideas/notes:**
- We could use this pagination with build-in commands: /help /chats etc
- CallbackqueryCommand could be somehow modified so that other commands could 'register' itself in it so it could redirect to commands based on data (just like in GenericmessageCommand conversation feature) - just prefix data with something like "c11_page1", must be short to fit into 64-character limit

noplanman:
> ...so user could use alphabetical pagination?

What exactly do you mean by this? Letters instead of numbers?

> button layouts

Brilliant, will see how to make this customisable :+1:

> We could use this pagination with build-in commands: /help /chats etc

Yep!

> CallbackqueryCommand could be somehow modified so that other commands could 'register' itself in it...

Exactly what I meant by `registerCallbackQueryHandler`, something like this:
```php
CallbackqueryCommand::registerCallbackQueryHandler('pagination_command', 'myCallbackQueryHandler');

function myCallbackQueryHandler($callback_query)
{
    ...
}
```

jacklul:
> ...so user could use alphabetical pagination?
>What exactly do you mean by this? Letters instead of numbers?

Yeah, sample use: sorting /chats by letters! No idea how hard this will be to achieve though!

noplanman:
> Would be nice if we could have few types of button layouts to choose from! Like even simple ones

Added with https://github.com/akalongman/php-telegram-bot/pull/438/commits/6983b2c0461311f07bc1d2c5c1c93dd1b3d09b1b, by hiding certain buttons using a blank label.

noplanman:
Will leave this open for discussion for the moment, but I'll be moving this whole feature to a separate repo, as an add-on.

Found this neat project that I'll be extending:
https://github.com/lartie/Telegram-Bot-Pagination

noplanman:
New project: https://github.com/php-telegram-bot/inline-keyboard-pagination/

PR #450, opened 2017-04-05
Update scope getCommandFromType

l3xx:
For the possibility of hierarchical inheritance


noplanman:
Thanks 👍

Are you using this at the moment, extending the Telegram class?

l3xx:
I override processUpdate method in my class and reused getCommandFromType

PR #456, opened 2017-04-07
No default commands

jacklul:
Do not load example user commands by default.
Reverted versions on 'dummy' user commands back to 1.0.0.

Unit Tests need a fix.


noplanman:
Tests are failing because there is no `EchoCommand` or `HelpCommand` any more.
As a next step to this PR, as soon as we've moved to an organisation (prod @akalongman), we should move all example commands to a dedicated Repo, no need to have dummy code in the core repo.
Will PR a fix 👍

jacklul:
Yeah, having a repository with example bot will be ideal

noplanman:
> Yeah, having a repository with example bot will be ideal

And on top of that, a repo filled with commands that can be reused.

lichtscheu:
+1 for Organisation!

akalongman:
Done: https://github.com/php-telegram-bot/library

tests/unit/Commands/CommandTestCase.php
```diff
@@ -37,7 +37,8 @@ class CommandTestCase extends TestCase
      */
     public function setUp()
     {
-        $this->telegram = new Telegram('apikey', 'testbot');
+        $this->telegram = new Telegram(self::$dummy_api_key, 'testbot');
+        $this->telegram->addCommandsPath(BASE_COMMANDS_PATH . '/UserCommands');
```
noplanman: This can be deleted, as there is no more `UserCommands` folder.

PR #457, opened 2017-04-07
Misc additions/changes

jacklul:
Add `getBotId()` method to `Telegram` object.
Regex extracting the id from the API key should be fine.

Rename `bot_name` variables (and similar) into `bot_username` to prevent confusion.

Unit Tests need a fix.


noplanman:
Tests are failing because the dummy API keys I used in the tests are bad, they don't have the correct format. So when your regex tries to extract the bot ID, there is none.
Making a fix PR right now, finished in a moment.

noplanman:
@akalongman @MBoretto 
You guys happy with these changes?

PR #464, opened 2017-04-13
Update DB.php

no-nevis:
Fix 'SQLSTATE[HY093]: Invalid parameter number' PDO error which caused of multiple parameters usage in one query.

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -367,12 +367,12 @@ class DB
                 INSERT INTO `' . TB_USER . '`
                 (`id`, `username`, `first_name`, `last_name`, `created_at`, `updated_at`)
                 VALUES
-                (:id, :username, :first_name, :last_name, :date, :date)
+                (:id, :username, :first_name, :last_name, :date, VALUES(`created_at`))
                 ON DUPLICATE KEY UPDATE
-                    `username`   = :username,
-                    `first_name` = :first_name,
-                    `last_name`  = :last_name,
-                    `updated_at` = :date
+                    `username`   = VALUES(`username`),
+                    `first_name` = VALUES(`first_name`),
+                    `last_name`  = VALUES(`last_name`),
+                    `updated_at` = VALUES(`created_at`)
             ');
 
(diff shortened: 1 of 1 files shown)


noplanman:
Thanks for your contribution!

What version of MySQL are you using, I have never encountered this issue.

no-nevis:
I'm using lumen framework. MySQL version: 5.1.73 (this error came from the pdo plugin itself, not mysql).

no-nevis:
why u dont merge it ?

noplanman:
Hi, sorry for the epic delay...

A cleaner solution in my view, is to simply rename the fields to `created_date` and `updated_date` respectively.

Also, your `ON DUPLICATE KEY` section saves the wrong value to `updated_at`, as it simply puts the value of `created_at` into the `updated_at` field, instead of the current timestamp.

If you could change it accordingly and do the same for all other instances of duplicate `:date` tokens in other queries, I'll happily merge 😃

noplanman:
@no-nevis Shall I go ahead and make the changes and merge?

no-nevis:
> Also, your ON DUPLICATE KEY section saves the wrong value to updated_at, as it simply puts the value of created_at into the updated_at field, instead of the current timestamp.
Fixed.

>A cleaner solution in my view, is to simply rename the fields to created_date and updated_date respectively.
The problem isn't only dates, but the other placeholders that used twice.

noplanman:
Hi @no-nevis 

I have taken the liberty to rebase your PR properly, to prevent the duplicate commits that snuck in.

Also, I've added another commit that renames the placeholders to make it more obvious and cleaner.

There is 1 query left that has duplicate (or rather triplicate) placeholders here:
https://github.com/no-nevis/core/blob/patch-1/src/DB.php#L1104

Is this one an issue for you too? Do you get the same error?

noplanman:
Thanks @no-nevis 😃👍

sana8237:
Delete

PR #466, opened 2017-04-15
Another small additions and changes

jacklul:
- Add isRunCommands()

Discussed here https://github.com/php-telegram-bot/core/issues/444
Couldnt find a better name for this function.

- Add optional parameter 'interval' to modify check interval for request limiter

By default there is 1 second delay between a check, which is fine but in some cases developer might want the delay to be slightly smaller (or higher!).

Important Notes:
Settings this to 0.5 will make the check every 0.5 second, this should improve bot resposiveness but can cause a slow down if database server is slow. 
Any value below 1 can cause requests actually reach Telegram API's limits - scenario where previous request was sent just before the next second pops in and another request is instantly served, if this happens too many times in a row Telegram API will issue a block, this is fine for small bursts.

- Prevent serving duplicate updates, throw exception instead

Issues with Telegram Bot API servers recently, attempt to prevent duplicated replies...

Also figured out how to get telegram_update's date:
This will select all records older than `2017-04-15 13:32:25`:
```
SELECT * FROM `telegram_update` WHERE
(`message_id` IS NOT NULL AND `message_id` IN (SELECT f.id FROM `message` f WHERE `date` < '2017-04-15 13:32:25')) OR 
(`edited_message_id` IS NOT NULL AND `edited_message_id` IN (SELECT f.id FROM `edited_message` f WHERE `edit_date` < '2017-04-15 13:32:25')) OR 
(`inline_query_id` IS NOT NULL AND `inline_query_id` IN (SELECT f.id FROM `inline_query` f WHERE `created_at` < '2017-04-15 13:32:25')) OR 
(`chosen_inline_result_id` IS NOT NULL AND `chosen_inline_result_id` IN (SELECT f.id FROM `chosen_inline_result` f WHERE `created_at` < '2017-04-15 13:32:25')) OR 
(`callback_query_id` IS NOT NULL AND `callback_query_id` IN (SELECT f.id FROM `callback_query` f WHERE `created_at` < '2017-04-15 13:32:25'))
```

(might be good idea to have a wiki article containing useful queries like this!?)

### DIFF
--- a/examples/cron.php
+++ b/examples/cron.php
@@ -64,7 +64,9 @@ try {
     //$telegram->setUploadPath('../Upload');
 
     // Requests Limiter (tries to prevent reaching Telegram API limits)
+    // Second argument are options
     $telegram->enableLimiter();
+    //$telegram->enableLimiter(['interval' => 0.5);
 
     // Run user selected commands
     $telegram->runCommands($commands);
--- a/examples/cron.php
+++ b/examples/cron.php
@@ -64,7 +64,7 @@ try {
     //$telegram->setUploadPath('../Upload');
 
     // Requests Limiter (tries to prevent reaching Telegram API limits)
-    // Second argument are options
+    // First argument are options
     $telegram->enableLimiter();
     //$telegram->enableLimiter(['interval' => 0.5);
 
--- a/src/DB.php
+++ b/src/DB.php
@@ -191,7 +191,7 @@ class DB
                 $sql .= ' WHERE `id` = :id';
             }
 
-            $sql.= ' ORDER BY `id` DESC';
+            $sql .= ' ORDER BY `id` DESC';
 
             if ($limit !== null) {
                 $sql .= ' LIMIT :limit';
--- a/src/DB.php
+++ b/src/DB.php
@@ -173,29 +173,37 @@ class DB
      * Fetch update(s) from DB
      *
      * @param int $limit Limit the number of updates to fetch
+     * @param int $id Check for unique update id
      *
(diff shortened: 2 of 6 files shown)


examples/cron.php
```diff
@@ -64,7 +64,9 @@
     //$telegram->setUploadPath('../Upload');
 
     // Requests Limiter (tries to prevent reaching Telegram API limits)
+    // Second argument are options
```
noplanman: First argument, it's the second argument for `setLimiter()`.

src/DB.php
```diff
@@ -173,29 +173,37 @@ public static function getPdo()
      * Fetch update(s) from DB
      *
      * @param int $limit Limit the number of updates to fetch
+     * @param int $id Check for unique update id
```
noplanman: Align description.

src/DB.php
```diff
@@ -474,6 +482,10 @@ public static function insertRequest(Update $update)
         $update_id   = $update->getUpdateId();
         $update_type = $update->getUpdateType();
 
+        if (count(self::selectTelegramUpdate(1, $update->getUpdateId())) == 1) {
```
noplanman: Use `===`.

src/Telegram.php
```diff
@@ -935,4 +947,14 @@ public function runCommands($commands)
             $this->executeCommand($this->update->getMessage()->getCommand());
         }
     }
+
+    /**
+     * Is this session initiated by runCommands()
+     *
+     * @return string
```
noplanman: Should be a bool.

jacklul:
Done!

PR #471, opened 2017-04-21
[minor] Forgotten $BOT_NAME => $BOT_USERNAME

Habetdin:
Minor example update with lil note bout $path


noplanman:
Thanks @Habetdin, taking care of the details, I like it 😃

PR #480, opened 2017-04-25
Bugfix: "Duplicate update received!" when not using DB

jacklul:


src/DB.php
```diff
@@ -482,7 +482,7 @@ public static function insertRequest(Update $update)
         $update_id   = $update->getUpdateId();
         $update_type = $update->getUpdateType();
 
-        if (count(self::selectTelegramUpdate(1, $update->getUpdateId())) === 1) {
+        if (self::isDbConnected() && count(self::selectTelegramUpdate(1, $update->getUpdateId())) === 1) {
```
noplanman: Best put that in an `if` at the beginning of the function, like the other ones have it too.

PR #490, opened 2017-05-03
Fix indexed column length and quotes in comments.

noplanman:
When using a MySQL (or MariaDB) server that does not have `innodb_large_prefix` enabled, indexed columns are limited to 191 characters when using `utf8mb4`.

Thus, when executing the query to create the `user` table, the following error occurs:
```
(HY000/1709): Index column size too large. The maximum column size is 767 bytes.
```

When importing the structure using `mysql` command or phpMyAdmin, all affected columns get set to `COLLATE utf8mb4_unicode_520_ci` on import, which prevents the error from popping up.

Since all columns at the moment are just set to `CHAR(255)`, I think it's better to adjust the column sizes to something more realistic.

For now, the `user.username` column is the only one affected.

Read more about [InnoDB limitations](https://dev.mysql.com/doc/refman/5.7/en/innodb-restrictions.html).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 ### Deprecated
 ### Removed
 ### Fixed
+- Indexed table columns now have a fitting length.
 ### Security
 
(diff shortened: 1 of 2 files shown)


noplanman:
Guys?

PR #512, opened 2017-06-03
Basic API 3.0 changes

jacklul:
Additions and changes:
- Added `deleteMessage` method
- Added `sendVideoNote` method
- Added `language_code` field for **User** entity
- Added `video_note` field for **Message** entity
- Added `gif_duration` field in **InlineQueryResultGif**
- Added `mpeg4_duration` field in **InlineQueryResultMpeg4Gif**
- Replaced the field `new_chat_member` with `new_chat_members` in **Message** entity

Not touched:
- **Payment Platform** stuff

What was tested and works:
- `language_code` thing
- `deleteMessage`
- `new_chat_members` DB field and `getNewChatMembers()`

What was NOT tested:
- **Video Notes** and DB's `video_note` field
- `sendVideoNote` method
- `gif_duration` and `mpeg4_duration` fields

Needs looking into:
- `new_chat_members` and related stuff (for example: if `Message->botAddedInChat()` works correctly)

There are probably LOTS of possible code improvements to do, I just simply had no time to write everything cleanly, sorry!

*Bot API 3.0 TODO issue: https://github.com/php-telegram-bot/core/issues/501*


noplanman:
@jacklul I think it would make more sense to save a comma separated list of IDs for the new `new_chat_members` column in the `message` table instead of the JSON objects. Otherwise there will be tons of redundant data in the DB.

Hmm, it's a bit annoying that the API is inconsistent with the `new_member` and `left_member`. Would be nice if they worked the same, i.e. also a `left_chat_members` field if multiple ones leave in 1 request.

chuv1:
But there is no way to remove multiple users in one request. So `left_member` is always single member, afaik.

noplanman:
@chuv1
Yes you're right, at the moment it's just 1 per request.
But maybe in the future it will be possible to remove multiple users in 1 go, just like adding them, at which point the API will have to change again.

noplanman:
@jacklul We'll need to start with migration scripts for changes to the DB structure, especially if it changes columns with existing data that should be kept.

For this, maybe the easiest would be to create the new column (`new_chat_members`) and then copy all data from `new_chat_member` to the new one, then delete the old one without FK constraints, otherwise it's a bit of a mess to rename and change type and FK all in one.

Have you given more thought to just saving a comma separated list of IDs instead of  all the User JSON objects?

jacklul:
@noplanman Sorry for not having much time lately, found little time to make this little change, is it okay?

src/DB.php
```diff
@@ -804,14 +804,19 @@ public static function insertMessageRequest(Message $message)
 
         //New and left chat member
         if (!empty($new_chat_members)) {
+            $new_chat_members_ids = '';
```
noplanman: Reminding you of using an array for this and then `implode`ing:
```php
$ids = [];
$ids[] = 123;
$ids[] = 456;
$ids_csv = implode(',', $ids);
```

If you want to keep it using a string, what you can change is:
```php
// this...
if (!empty($new_chat_members_ids)) {

// ...to this
if ($new_chat_members_ids) {
```

As it's the same thing.

noplanman:
@jacklul I've added a simple 2-line script that creates the new `new_chat_members` column and copies the user ID over from `new_chat_member`.

I'm not deleting the old column as some users may be relying on that data still, until they make the switch. So at some point when we redo the database, we'll just not export it when setting up the new structure. You think that's ok?

Issue #518, opened 2017-06-04
Chat ID wont get properly saved


no-nevis:
### Required Information

- Operating system: Shared Hosting
- PHP version:  5.6
- PHP Telegram Bot version: 0.44.1
- Using MySQL database: yes
- MySQL version: 5.1.73
- Update Method: Webhook
- Self-signed certificate: no
- RAW update (if available):

### Expected behaviour

Script should properly save group ids in the mysql.
### Actual behaviour

I can't get a proper chat id from groups. for example, I got this from a group: -403673758
And when i try sending a message to this chat, nothing happens and i got that real chat id is this:
-1001128098546
### Steps to reproduce

This happened recently. i don't know.

no-nevis:
But the PDO engine of PHP can insert only the max integer value of PHP, i.e. 2147483647.
I found this on a stack overflow question :|

no-nevis:
I fixed it :|
By changing lines in DB.php:
```
if ($migrate_to_chat_id) {
                $chat_type = 'supergroup';

                $sth->bindParam(':id', $migrate_to_chat_id, PDO::PARAM_INT);
                $sth->bindParam(':oldid', $chat_id, PDO::PARAM_INT);
            } else {
                $sth->bindParam(':id', $chat_id, PDO::PARAM_INT);
                $sth->bindParam(':oldid', $migrate_to_chat_id, PDO::PARAM_INT);
            }
```
To:
```
if ($migrate_to_chat_id) {
                $chat_type = 'supergroup';

                $sth->bindParam(':id', $migrate_to_chat_id, PDO::PARAM_STR);
                $sth->bindParam(':oldid', $chat_id, PDO::PARAM_STR);
            } else {
                $sth->bindParam(':id', $chat_id, PDO::PARAM_STR);
                $sth->bindParam(':oldid', $migrate_to_chat_id, PDO::PARAM_STR);
            }
```
PDO cannot bind bigint.

noplanman:
Probably makes sense to save all IDs as strings in DB to prevent this and any further issues as ID will inevitably get bigger.

https://stackoverflow.com/questions/11938384/how-do-i-store-a-bigint-in-mysql-using-pdo

@jacklul @MBoretto @akalongman 
What do you think about this? Seems to have no disadvantage and makes it work properly.

PR #519, opened 2017-06-04
Fix issue #518 - Properly Save Chat IDs

no-nevis:
Fixing #518 issue, happening because of PDO not able to save large ints.


no-nevis:
Oops, wrong

PR #523, opened 2017-06-05
Fix case when handling channel messages (request limiter)

jacklul:
fixes https://github.com/php-telegram-bot/core/issues/509#issuecomment-306137865 i guess?

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -1064,8 +1064,8 @@ class Request
                     $requests = DB::getTelegramRequestCount($chat_id, $inline_message_id);
 
                     if ($requests['LIMIT_PER_SEC'] == 0     // No more than one message per second inside a particular chat
-                        && ((($chat_id > 0 || $inline_message_id) && $requests['LIMIT_PER_SEC_ALL'] < 30)       // No more than 30 messages per second globally
-                        || ($chat_id < 0 && $requests['LIMIT_PER_MINUTE'] < 20))        // No more than 20 messages per minute in groups and channels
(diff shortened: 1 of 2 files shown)


src/Request.php
```diff
@@ -1064,8 +1064,8 @@ private static function limitTelegramRequests($action, array $data = [])
                     $requests = DB::getTelegramRequestCount($chat_id, $inline_message_id);
 
                     if ($requests['LIMIT_PER_SEC'] == 0     // No more than one message per second inside a particular chat
```
noplanman: I think it would make this query easier to understand, if each part was saved in a separate variable before the `if`.

src/Request.php
```diff
@@ -1076,10 +1076,11 @@ private static function limitTelegramRequests($action, array $data = [])
 
                     $requests = DB::getTelegramRequestCount($chat_id, $inline_message_id);
 
-                    if ($requests['LIMIT_PER_SEC'] == 0     // No more than one message per second inside a particular chat
```
noplanman: @jacklul I'm looking at this again, as I'm setting up to release 0.45.0, and I really don't think that the `if` statement does the same thing as before.
Is this intentional to fix something?

jacklul: Intentional, `LIMIT_PER_SEC_ALL` should also apply to groups, it didn't before.

At some point the limiter will have to be redone, there is more limits than they put in the documentation (editing has different limits than sending).

noplanman: Ok, perfect! Will merge then 👍

PR #524, opened 2017-06-05
Allow selecting channels with selectChats() and sendToActiveChats()

jacklul:
This can be particulary useful now since bots can receive channel messages.

I hope I didn't miss anything, I guess travis will tell...

### DIFF
--- a/src/Commands/AdminCommands/ChatsCommand.php
+++ b/src/Commands/AdminCommands/ChatsCommand.php
@@ -58,6 +58,7 @@ class ChatsCommand extends AdminCommand
         $results = DB::selectChats(
             true, //Select groups (group chat)
             true, //Select supergroups (super group chat)
+            true, //Select channels
             true, //Select users (single chat)
             null, //'yyyy-mm-dd hh:mm:ss' date range from
             null, //'yyyy-mm-dd hh:mm:ss' date range to
@@ -68,6 +69,7 @@ class ChatsCommand extends AdminCommand
         $user_chats        = 0;
         $group_chats       = 0;
         $super_group_chats = 0;
+        $channel_chats     = 0;
 
         if ($text === '') {
             $text_back = '';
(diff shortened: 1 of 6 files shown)


src/Commands/AdminCommands/ChatsCommand.php
```diff
@@ -58,6 +58,7 @@ public function execute()
         $results = DB::selectChats(
```
noplanman: This is screaming for bit flags / bitmask:
https://stackoverflow.com/questions/11880360/how-to-implement-a-bitmask-in-php

Having so many boolean parameters is just waaay confusing 😖

jacklul: Well, I think I have no clue how to do this in a clean way, I would have to declare those somewhere, but where to make them accessible at all time?

noplanman: An option could be to just pop them into the DB class:
```php
const SINGLE_CHAT = 1;
const GROUP_CHAT = 2;
const SUPERGROUP_CHAT = 4;
const CHANNEL = 8;
```
But this may seem a bit confusing 😕 

Probably makes sense to do the same here like with other parameters, passing the values as an array (as described further down from my link above: https://stackoverflow.com/a/29089798):
```php
$results = DB::selectChats([
    'groups'       => true,
    'supergroups'  => true,
    'channels'     => true,
    'single_chats' => true, // or 'user_chats', or just 'single'?
    'from'         => 'yyyy-mm-dd hh:mm:ss', // or better, use DateTimeImmutable objects
    'to'           => 'yyyy-mm-dd hh:mm:ss',
]);
```

noplanman: @jacklul Have made a PR to your PR that implements the solution with the array, please check it out when you have a moment, thanks.

PR #540, opened 2017-06-19
Callbacks can be added to be executed when callback queries are called.

noplanman:
This allows adding callback hooks to CallbackqueryCommand:

e.g. in `hook.php` when setting up the lib:
```php
CallbackqueryCommand::addCallbackHandler(function (CallbackQuery $query) {
    if ($query->getFrom()->getUsername() === 'noplanman') {
        // Do something with this callback, but only when requested by @noplanman
    }
});
```

Any number of callback handlers can be added like this. All of them get run when a callback is being executed, so each callback function needs to do its own checks to handle the proper calls.

@jacklul I've been thinking of ways to extend this, but I'd like to keep it as simple as possible for now, and I feel this is already quite powerful in itself and allows users to very easily add their own custom code to callbacks. What do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 
 ## [Unreleased]
 ### Added
+- Callbacks can be added to be executed when callback queries are called.
 ### Changed
 ### Deprecated
 ### Removed
--- a/src/Commands/SystemCommands/CallbackqueryCommand.php
+++ b/src/Commands/SystemCommands/CallbackqueryCommand.php
@@ -19,6 +19,11 @@ use Longman\TelegramBot\Request;
 class CallbackqueryCommand extends SystemCommand
 {
     /**
+     * @var callable[]
+     */
+    protected static $callbacks = [];
+
+    /**
      * @var string
      */
     protected $name = 'callbackquery';
@@ -46,6 +51,21 @@ class CallbackqueryCommand extends SystemCommand
         //$query_id       = $callback_query->getId();
         //$query_data     = $callback_query->getData();
 
+        // Call all registered callbacks.
+        foreach (self::$callbacks as $callback) {
+            $callback($this->getUpdate()->getCallbackQuery());
+        }
+
         return Request::answerCallbackQuery(['callback_query_id' => $this->getUpdate()->getCallbackQuery()->getId()]);
+    }
+
+    /**
+     * Add a new callback handler for callback queries.
+     *
+     * @param $callback
+     */
+    public static function addCallbackHandler($callback)
+    {
+        self::$callbacks[] = $callback;
     }
 }


MyZik:
Hi, @noplanman ! If I use that in my hook.php, or in my random command I get an error:

`PHP Fatal error:  Cannot declare class Longman\\TelegramBot\\Commands\\SystemCommands\\CallbackqueryCommand, because the name is already in use in /var/www/www-root/data/www/telebotpro.montanabank.pw/Commands/SystemCommands/CallbackqueryCommand.php on line 0`

noplanman:
You probably also have a custom `CallbackqueryCommand.php` in your custom commands folder, right?

MyZik:
@noplanman yep. In /Commands/SystemCommands/ I have a CallbackqueryCommand.php
Is it unright?

noplanman:
That's a custom one, right? You'll need to add it to your composer.json so that it gets loaded before the library one, like this:
```json
    "autoload": {
        "classmap": [
            "your/custom/comands/folder"
        ]
    }
```
Then, do a `composer dump-autoload` to reload the composer files.

Not sure this is the best solution, but it works for me.

MyZik:
@noplanman thanks!
But I have some problem. When I press the button, nothing happens :(
Logs are empty...

Here is my hook.php
```php
<?php

...

CallbackqueryCommand::addCallbackHandler(function (CallbackQuery $query) {
        $callback_data = $query->getData();

        /**
         * Callback for "Company" section
         */

        if (strpos($callback_data, 'about_') !== 0) {
            return Request::emptyResponse();
        }

        // Get the real category from the callback_data.
        $current = substr($callback_data, strlen('about_'));
        switch ($current) {
            /**
             * About us
             */
            case 'us':
                $out_text = 'We're cool!';
                break;
                
            /**
             * Our team
             */
            case 'team':
                $out_text = 'John Doe - dev';
                break;

            /**
             * Contacts
             */
            case 'contacts':
                $out_text = 'Contacts us! test@test.com';
                break;

            /**
             * Cooperate
             */
            case 'cooperation':
                $out_text = 'We love you too!';
                break;

            /**
             * If smthng horrorable cares
             */
            default:
                $out_text = 'Default text*';
            }

            return Request::editMessageText([
                'chat_id' => $query->getMessage()->getChat()->getId(),
                'message_id' => $query->getMessage()->getMessageId(),
                'text' => $out_text,
                'reply_markup' => AboutCommand::getMyInlineKeyboard(),
            ]);
    });
```

And here is my `AboutCommand::getMyInlineKeyboard()`
```php
public static function getMyInlineKeyboard()
    {
        $keyboard = new InlineKeyboard(
            [
                ['text' => About us', 'callback_data' => 'about_us'],
                ['text' => 'Our team', 'callback_data' => 'about_team'],
            ],
            [
                ['text' => 'Contacts', 'callback_data' => 'about_contacts'],
                ['text' => 'Cooperation', 'callback_data' => 'about_cooperation'],
            ]
        );
        return $keyboard;
    }
```

PR #554, opened 2017-07-03
fix(structure.sql): Remove wrong character in database structure

nasserghiasi:
`new_chat_members` column name to `new_chat_member`


jacklul:
No, it's correct as per API changes: https://core.telegram.org/bots/api#may-18-2017

> Replaced the field new_chat_member in Message with new_chat_members (the old field will still be available for a while for compatibility purposes).

PR #565, opened 2017-07-18
Scheduler patch

jacklul:
Fixes two issues.

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -918,8 +918,10 @@ class Telegram
             $bot_username = $this->getBotUsername();
         }
 
+        if (!$this->isAdmin($bot_id)) {
+            $this->enableAdmin($bot_id);    // Give bot access to admin commands
+        }
 
-        $this->enableAdmin($bot_id);    // Give bot access to admin commands
         $this->getCommandsList();       // Load full commands list
 
         foreach ($commands as $command) {
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -520,10 +520,10 @@ class Telegram
      */
     public function enableAdmin($admin_id)
     {
-        if (is_int($admin_id) && $admin_id > 0 && !in_array($admin_id, $this->admins_list, true)) {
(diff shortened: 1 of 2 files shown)


src/Telegram.php
```diff
@@ -904,17 +904,24 @@ public function runCommands($commands)
...
+        } else {
+            $bot_id       = $this->getBotId();
+            $bot_name     = $this->getBotUsername();
+            $bot_username = $this->getBotUsername();
         }
 
-        $bot_id       = $result->getId();
-        $bot_name     = $result->getFirstName();
-        $bot_username = $result->getUsername();
+        if (!$this->isAdmin($bot_id)) {
```
noplanman: Instead of the `isAdmin` check, just cast bot id to `int`:
`$this->enableAdmin((int) $bot_id);`

jacklul: I don't think this is an issue here, since it works in first loop then starts failing, isnt there a check in addadmin to return false when Id is already admin? This is causing this.

noplanman: Yes, [there is a check](https://github.com/php-telegram-bot/core/blob/0.46.0/src/Telegram.php#L523), but the problem is, that the passed ID must be an integer, and it's probably being passed as a string.

jacklul: Actually, it's not being passed as string, I will add `(int)` just to be sure but it wasn't the case!

noplanman: So weird, I'm wondering why it's failing then 🤔
Are you able to debug this?

noplanman: Oh, silly me! Of course, the else is totally broken... It should just skip if the id has already been added. Will fix that in this PR 👍

noplanman: Ok, have a look and test this one please: https://github.com/php-telegram-bot/core/pull/565/commits/d806cde56c717b933c691fc916ea97350d9b8ac4

jacklul: Yep, it works properly now!

PR #566, opened 2017-07-19
Cleanup without database...

jacklul:
Prevent running /cleanup when there is no DB connection...

_I don't think it will clean up the universe in that scenario._

### DIFF
--- a/src/Commands/AdminCommands/CleanupCommand.php
+++ b/src/Commands/AdminCommands/CleanupCommand.php
@@ -351,19 +351,13 @@ class CleanupCommand extends AdminCommand
     public function execute()
     {
         $message = $this->getMessage();
-        $chat_id = $message->getChat()->getId();
+        $user_id = $message->getFrom()->getId();
         $text    = $message->getText(true);
 
         $data = [
-            'chat_id'    => $chat_id,
+            'chat_id'    => $user_id,
             'parse_mode' => 'Markdown',
         ];
-
-        if (!$message->getChat()->isPrivateChat()) {
-            $data['text'] = '/cleanup command is only available in a private chat.';
-
-            return Request::sendMessage($data);
-        }
 
         $settings = $this->getSettings($text);
         $queries  = $this->getQueries($settings);
--- a/src/Commands/AdminCommands/CleanupCommand.php
+++ b/src/Commands/AdminCommands/CleanupCommand.php
@@ -58,6 +58,11 @@ class CleanupCommand extends AdminCommand
     protected $version = '1.0.0';
 
     /**
+     * @var bool
+     */
+    protected $need_mysql = true;
+
+    /**
      * Default tables to clean, cleaning 'chat', 'user' and 'user_chat' by default is bad practice!
      *
      * @var array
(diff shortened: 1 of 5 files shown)


src/Commands/AdminCommands/CleanupCommand.php
```diff
@@ -327,7 +351,7 @@ private function getQueries($settings)
     public function execute()
     {
         $message = $this->getMessage();
-        $chat_id = $message->getFrom()->getId();
+        $chat_id = $message->getChat()->getId();
```
noplanman: Thought it makes sense to keep this private, as not to expose any info in a group.
Mind explaining why the change to `getChat` now?

src/Commands/AdminCommands/CleanupCommand.php
```diff
@@ -319,6 +324,25 @@ private function getQueries($settings)
...
+     * @return \Longman\TelegramBot\Entities\ServerResponse
+     */
+    public function executeNoDb()
+    {
+        $message = $this->getMessage();
+        $chat_id = $message->getChat()->getId();
+
+        $data = [
+            'chat_id'    => $chat_id,
+            'parse_mode' => 'Markdown',
```
noplanman: Not that it makes any difference, but no need for `parse_mode` here.

src/Commands/AdminCommands/CleanupCommand.php
jacklul: Because... `getFrom()->getId()` would return id of the user who called this command, so the bot would reply to the private chat with that user.

src/Commands/AdminCommands/CleanupCommand.php
jacklul: Now it does!

src/Commands/AdminCommands/CleanupCommand.php
noplanman: Yes, I know that, which is why I thought you put that there in the first place :smile:
Because with `getChat`, it posts any error or the final summary visible for anyone in a chat to read.
IMHO, maintenance stuff should be managed privately.

src/Commands/AdminCommands/CleanupCommand.php
noplanman: Haha, you crack me up... 😉

src/Commands/AdminCommands/CleanupCommand.php
jacklul: Very rare case would be an admin who didnt message his bot in private chat yet.
All admins commands should get some kind of change like this, maybe we could decide on something and do it in this PR?

noplanman: Agreed, it's a rare case, but it can happen by mistake. Using `getFrom()` is a fail-safe for that, as the Chat ID would be the same as the From ID anyway in a private chat, so it doesn't make a difference there.

Haven't tested this yet, but can a bot delete a message from a user when it's admin in a group/channel? Then, we could even go as far as to offer the option of deleting a "by mistake" command execution.

What do you think?

jacklul: Yes, bot can delete user messages when he has the permissions to do so.

That would be a nice solution!

noplanman: Ok, then I think that's what we should go for 👍
Will play around with this when I have a moment.

PR #573, opened 2017-07-25
fix get command from type 'new_chat_members'

lensert:


noplanman:
Aah, nice catch! 👌 

There's always one that thinks he can hide and get away with it 😄

Issue #112, opened 2016-02-28
DB schema update scripts


noplanman:
Having update scripts for the database schema will make it a lot easier to make sure that the correct structure is being used and no errors happen.

For this we need to save the current database schema version somewhere in the database, which we can do by adding an extra table, e.g. `options`. For the future, we could also use this table to save options, like the bot name and API key for example.

MBoretto:
Will come with monolog feature
https://github.com/akalongman/php-telegram-bot/issues/167

noplanman:
@MBoretto 
This issue here is about DB update scripts that simply update the schema depending on the library version. It's different to the "import all data" that you have implemented with the Monolog PR.

The problem is, that users who haven't logged all the requests, will have the same problem every time the schema is changed, because they can't just reimport everything 😕

MBoretto:
ok

noplanman:
Right, I'm at a stage where foreign key constraints are not my best friends 😄 

Check out my work so far here:
https://github.com/noplanman/php-telegram-bot/tree/db_schema_update/utils/db-schema-update

This script automatically:
- fetches all tags from the live repo
- imports and dumps each structure.sql
- compares them for changes
- creates a migration script for any changes
- applies the migration to a test database

It starts at version 0.16.0, as anything before then is very shaky and unstable.
This migration works perfectly when starting with a DB at version 0.16.0, but has issues when starting the migration at any version in between.

Reason for this: *drumroll...* **key constraints**!

So I'm in the process of trying to figure out how best to solve this. The problem is, that each time a foreign key is set without explicitly setting a constraint name (via `ADD CONSTRAINT \`<this-name!>\` FOREIGN KEY (\`user_id\`) REFERENCES \`user\` (\`id\`);`), a new one with an auto-incremented name is generated (`message_ibfk_1`, `message_ibfk_2`, etc.).
Because of this, as the migration proceeds, some get removed and re-added, thus making the number go up. So when starting the process somewhere in between, the numbers are wrong, causing issues (e.g. `Duplicate foreign key constraint name 'tmp/message_ibfk_X'`).

I'm really working towards a fully automated way of handling this, but it's becoming more and more complex and I'm not sure it's worth it 😕 

I might just leave this as is and make dedicated migration scripts manually...

As for any new structure changes, with each change, a migration script is a **MUST**, to prevent this mess.

What do you think? Any thoughts or ideas?

Sogl:
What about adding a small ORM and stop using FK?

[...]

PR #590, opened 2017-08-13
Add laravel migrations

mithek:
This pull requests contains a laravel service provider to import the migrations.
It also contains the migrations based of the structure.sql

This pull request should make it easier for laravel programmers to use this package with migrations.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,8 @@
         "ext-curl": "*",
         "ext-mbstring": "*",
         "monolog/monolog": "^1.22",
-        "guzzlehttp/guzzle": "^6.2"
+        "guzzlehttp/guzzle": "^6.2",
+        "illuminate/support": "5.0.*|5.1.*|5.2.*|5.3.*|5.4.*|5.5.*"
     },
     "require-dev": {
         "phpunit/phpunit": "^4.8|^5.7|^6.1",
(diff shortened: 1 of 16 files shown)


jacklul:
It needs to be integrated with custom table prefix, you assumed usage of `telegram_` prefix which not everyone will be using!

mithek:
@jacklul I've moved the prefix to a config file. I also added the config file within the laravel service provider.  I'm not sure how I can add my latest commit to the PR.

noplanman:
Hi @mithek 

First of all, thanks for the PR and your idea / implementation!
Sorry for the delayed response from my side.

The topic of migrations has been raised and discussed a long while ago here: #112 
Since 0.45.0 onward (which is still very recent as of writing), simple SQL update scripts are the easiest to manage, without adding extra complexity / dependencies to the project.

We are well aware that a new DB layer / ORM should be implemented, but at the moment there just aren't enough resources for this 😕 

There are various projects out there that manage DB migrations, so we'd need to decide which approach to take when the time comes.

For the above reasons, I'll close this PR for now, but we'll definitely get back to it should we take the Laravel migration approach!

It would be best to first post an issue with an idea that can be discussed, before implementing something, just to keep unused code to a super-minimum 😉

jacklul:
Just want to add this topic might come back once we decide to support multiple database drivers.

PR #627, opened 2017-09-10
Develop

Maxiride:
fixed getUpdates link the in the index. An "s" was missing.


Maxiride:
choosed the wrong branch ^^

PR #628, opened 2017-09-10
Update README.md

Maxiride:
Fixed "s" missing from the broken getUpdate link in the index.


noplanman:
@Maxiride Bonus points if you change it everywhere 😉 
(like the title just next to it)

Also, there are some others littered around, do a quick search please, thanks!

noplanman:
Thanks @Maxiride 🎉

PR #641, opened 2017-09-16
No new feature, No bug fix

imanghafoori1:
Only code clean up for better code quality


noplanman:
@imanghafoori1 Thanks for your contribution!

I have changed a few parts of the code and made use of existing helpers which you probably didn't know about. Now these functions are cleaner, thank you 👍

PR #647, opened 2017-09-20
Add missing Payment properties to Message entity

noplanman:
Also add missing type checks, removing some unnecessary ones.

@jacklul Just to be safe, `Telegram.php:423` really isn't required, right? It's literally an array of all other types except for `message`, which is checked in the other `if`. There can't be any other value there anyway, right?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 
 ## [Unreleased]
 ### Added
+- Finish implementing payments, adding all missing type checks and docblock methods.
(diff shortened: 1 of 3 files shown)


jacklul:
> @jacklul Just to be safe, Telegram.php:423 really isn't required, right? It's literally an array of all other types except for message, which is checked in the other if. There can't be any other value there anyway, right?

Correct, "there shouldn't be any other value" - e.g. Telegram releases new update type.

noplanman:
> Telegram releases new update type

True. I think it's safe leave it like this and let any new type follow the default behaviour, right?
Only exception is `message`.

PR #654, opened 2017-09-23
Scrutinizer improvements

noplanman:
Some improvements and fixes found by Scrutinizer, reducing the timeout for external code coverage to speed up manual triggers.

Fix return value of `Message::getCommand()` if message is text only. Now returns `null` instead of `false`.

### DIFF
--- a/src/DB.php
+++ b/src/DB.php
@@ -452,7 +452,7 @@ class DB
             $chat_id   = $chat->getId();
             $chat_type = $chat->getType();
 
-            if ($migrate_to_chat_id) {
+            if ($migrate_to_chat_id !== null) {
                 $chat_type = 'supergroup';
 
                 $sth->bindValue(':id', $migrate_to_chat_id);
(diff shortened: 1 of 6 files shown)


src/DB.php
```diff
@@ -872,7 +872,7 @@ public static function insertMessageRequest(Message $message)
             $sth->bindValue(':forward_date', $forward_date);
 
             $reply_to_chat_id = null;
-            if ($reply_to_message_id) {
+            if ($reply_to_message_id !== null) {
```
Pub4Game: `!is_null()`

noplanman: @Pub4Game Welcome to the "every millisecond counts" game 😹 
`is_null()` is slower, as it adds a function call.

Using PHP5, it can be up to 10x slower.
Look here and give the code example a try:
https://hakre.wordpress.com/2011/03/29/php-is_null-vs-null/

PHP7+ fixes this, making `is_null()` a bit quicker even sometimes.

Pub4Game: I think we live in a time when php5 needs to go out of our lives... 😃

noplanman: Haha, for sure! When 7.2 is officially out, i really hope more people start dropping 5.

PR #675, opened 2017-10-11
Bot api 3.4

### LINKED ISSUE #674: Bot API 3.4
https://core.telegram.org/bots/api-changelog#october-11-2017

- [x] **Live Locations**: Add new field `live_period` to the `sendLocation` method and the `editMessageLiveLocation` and `stopMessageLiveLocation` methods.
- [x] Add new `setChatStickerSet` and `deleteChatStickerSet` methods to manage group sticker sets.
- [x] The `getChat` request now returns the group's sticker set for supergroups if available.
- [x] Add new field `caption_entities` in `Message` entity, to receive entities from media captions.

noplanman:
This implements latest additions for Bot API 3.4.

Closes #674


src/Request.php
```diff
@@ -654,6 +664,8 @@ private static function limitTelegramRequests($action, array $data = [])
                 'editMessageReplyMarkup',
                 'setChatTitle',
                 'setChatDescription',
+                'setChatStickerSet',
```
noplanman: @jacklul What exactly are the criteria for having methods in the limiter?
I see that some methods are listed, some not.

jacklul: API docs stated that only methods sending something or editing something in the chats are affected by limits.

noplanman: Ok, so these Sticker Set methods should be limited then 👍

PR #707, opened 2017-11-19
Add new sendMediaGroup method

zahhar:
sendMediaGroup method was missing, see https://core.telegram.org/bots/api#sendmediagroup

ISSUE: https://github.com/php-telegram-bot/core/issues/706


mehdikhody:
I upload with `\CURLFILE` my photos, videos and ... .
but in new version it's not possible to upload files with `\CURLFILE` in InputMedia.

mehdikhody:
use this.
https://github.com/botstan/API#inputmedia

noplanman:
Great @zahhar, thanks for this. Will check it out tomorrow and add any other missing API additions too 👍

PR #733, opened 2017-12-28
Log api.telegram response errors

VoDmAl:
Telegram has very strong size restriction for Reply Button callback_data. It is very helpful to see response error in error log. 

I do not create issue as this is very small improvement.

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -360,6 +360,7 @@ class Request
                 TelegramLog::update($result);
             }
         } catch (RequestException $e) {
+            TelegramLog::error($e);
             $result = ($e->getResponse()) ? (string) $e->getResponse()->getBody() : '';
         } finally {
             //Logging verbose debug output


jacklul:
Your solution will just flood log file.

When `Request` fails (you can check it by `$result->isOk()`) you cannot get error description with `$result->getDescription()` when you need to handle failures.

VoDmAl:
@jacklul, do you propose to change from error to debug? It is OK for me. May be it is even better. As now in debug log I see only this useless information without useful details:

```
* upload completely sent off: 2458 out of 2458 bytes
< HTTP/1.1 400 Bad Request
* Server nginx/1.12.2 is not blacklisted
< Server: nginx/1.12.2
< Date: Wed, 27 Dec 2017 18:01:50 GMT
< Content-Type: application/json
< Content-Length: 78
< Connection: keep-alive
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Content-Length,Content-Type,Date,Server,Connection
<
* Connection #0 to host api.telegram.org left intact
```

Inconsistency in error handling is an other issue of this great library. Somewhere try/catch, somewhere isOk.

noplanman:
> Inconsistency in error handling is an other issue of this great library. Somewhere try/catch, somewhere isOk.

I totally agree with you that the whole error handling and logging needs a revamp.
As @jacklul mentions though, this will flood the logs even more, but what is needed is more precise error handling and message outputs, to allow easier pinpointing of issues, instead of a flood of messages.

I'll leave this PR open for the moment as a statement of "get the logging and error handling sorted already!".

noplanman:
Something like this has been implemented in #1089

PR #774, opened 2018-02-09
Update scripts lines to work on both Unix and Windows

jacklul:
This shouldn't break anything.
Tested it in Windows command line and linux subsystem.

Currently on windows command line it doesn't seem to execute:
```
>composer check-code
> ./vendor/bin/phpcs --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150
'.' is not recognized as an internal or external command,
operable program or batch file.
Script ./vendor/bin/phpcs --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150 handling the check-code event returned with error code 1
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,13 +41,13 @@
     },
     "scripts": {
         "check-code": [
-            "./vendor/bin/phpcs --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150"
+            "\"vendor/bin/phpcs\" --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150"
         ],
         "test": [
-            "./vendor/bin/phpunit"
+            "\"vendor/bin/phpunit\""
         ],
         "test-cov": [
-            "./vendor/bin/phpunit --coverage-clover build/logs/clover.xml"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -41,13 +41,13 @@
     },
     "scripts": {
         "check-code": [
-            "./vendor/bin/phpcs --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150"
+            "\"vendor/bin/phpcs\" --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150"
```
noplanman: @jacklul Have you also tried it without the escaped double-quotes?
Just: `"vendor/bin/phpcs ..."`

jacklul: Yep! Doesn't work either!

```
>composer check-code
> vendor/bin/phpcs --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150
'vendor' is not recognized as an internal or external command,
operable program or batch file.
Script vendor/bin/phpcs --standard=phpcs.xml -snp --encoding=utf-8 src/ tests/ --report-width=150 handling the check-code event returned with error code 1
```

noplanman: Uff, windows... 😁

PR #776, opened 2018-02-11
[WIP] Prevent system commands from being called by the user directly

### LINKED ISSUE #772: User calling System commands fails
Because System commands can be called in the same way that User and Admin commands are called, simply executing a System command from the chat shouldn't be possible.

A quick fix would be simply edit each System command to see if it has been called by a User or the bot itself.
The `/start` command would need special treatment, because that should still work as a normal User command too.

As for the others, maybe it would make sense to do the check further up instead of inside the command itself.

What do you guys think @php-telegram-bot/developers

noplanman:
System commands can't be called by users directly any more (e.g. `/inlinequery`, `/newchatmembers`, etc.)

WIP part of this, is that `/generic` and `/genericmessage` can still be called directly by the user, as those are required as fallback commands. If overridden by the bot developer, this needs to be taken in to account to prevent security or consistency issues!
Didn't find a nicer implementation to fix as it is now.
@jacklul I'm very open to better suggestions and ideas 👍 

❗️ Also, `StartCommand` is now a `UserCommand`, so all custom implementations need to be adjusted, as all `/start` commands defined as `SystemCommand` will not be executed any more!

fixes #772


noplanman:
@jacklul I can't seem to reproduce the error you're getting... It works for me 😕 

Are you using a custom generic and/or genericmessage command?

Regarding the command rewrite, absolutely it needs a remake, which will be very **NOT** backwards compatible 😁

jacklul:
Both `/generic` and `/genericmessage` are custom in my case.
I also have commands dir pointing to example-bot's commands set!

noplanman:
@jacklul That error has only started happening for you with this PR?

jacklul:
Yep

NCLS96:
Why you didn't use 'usage' property of commands to prevent SystemCommands from Calling by User  (If usage not null or command object has 'usage' property, this command can be called by the user directyl)? What is the purpose of 'usage' properties in commands's class?

noplanman:
@php-telegram-bot/developers Mind having another look at this please?

noplanman:
Superseded by #970

PR #813, opened 2018-04-06
Prevent redefine constant BASE_PATH

customsoft:
Add defined for not redefine constant BASE_PATH


noplanman:
Right, thanks for this PR :+1: 

Will rename the constants and make the appropriate checks for the others too.

akalongman:
Ideally we should not use constants at all, but for current version its ok

PR #815, opened 2018-04-10
Develop

p1ratrulezzz:


akalongman:
@p1ratrulezzz what is this??

noplanman:
@p1ratrulezzz May I ask what your goal is with all these changes and your fork?

Maybe an alternative for you would be to contribute to this original library instead of making a copy and putting your name on it?

p1ratrulezzz:
Ah this was very specific case, I needed to run some commands before previous commands. I'm not sure if I can provide a good solution for now because still investigating into the code. If I come up with idea I'll send a correct pull request of course

noplanman:
Feel free to open an issue with your specific use-case :+1: 

If it's a general problem, it would be best to have a proper solution in case other users are experiencing the same thing.

Issue #864, opened 2018-07-19
Send chat action automatically


smoqadam:
Sending chat action automatically by calling any `send...` command. for example the following command send `typing...` chat action:

```php
$result = Request::sendMessage([
    'chat_id' => $chat_id,
   'sendChatAction' => true,
    'text'   => 'test',
]);
```

noplanman:
@smoqadam I like this idea!
We'd have to establish a safe way of passing the value though, as the `$data` array is what gets sent to the Telegram API. So we'd either have to make sure that it is a key that will *never* be used by Telegram, or (and maybe better), have an extra parameter for `$extras` like this.

@php-telegram-bot/developers 
What do you guys think about this?
Do you use the chat actions often enough to make this a necessary / 'makes sense' addition?

jacklul:
As API says it should only be used if bot is about to do something that is gonna take some time, it should definitely not be default behaviour.

noplanman:
@jacklul It isn't default! That's why it would be passed as an extra parameter :+1:

smoqadam:
@jacklul @noplanman 
you mean like this?
```php
Request::sendMessage([
    'chat_id' => $chat->getId(),
    'text' => 'test'
], ['sendChatActiob'=>true]);
```

noplanman:
@chuv1 I like things that make it easier for the user, but I agree with you that it's a question of adding a single line to send the chat action separately.

I do see the advantage of the parameter though, to easily define a long operation with a boolean and have the chat action be managed automatically, without any extra `if` or anything.

@smoqadam Yes, something like that was in my head. Haven't checked to see if that makes complete sense, but it should do the trick i think.
Before coding more, let's first establish that it's something that would be merged in the first place :blush:

jacklul:
Actually, @chuv1 is right.

noplanman:
Aah... Of course... I read his comment wrongly.
Of course it should be sent as the request comes in, not out.

noplanman:
So all ok to close this then?

@smoqadam Thanks a lot for the idea though! Keep them coming :+1:

smoqadam:
@noplanman no problem.

 I'm glad I to be helpful if I can

PR #865, opened 2018-07-19
#864 Send chat action automatically

smoqadam:
#865 Send chat action automatically


noplanman:
@jacklul I totally agree with you on this.
Note my comment here: https://github.com/php-telegram-bot/core/issues/864#issuecomment-406253485

src/Request.php
```diff
@@ -717,4 +719,51 @@ private static function limitTelegramRequests($action, array $data = [])
...
+
+
+    private static function sendChatActionIfNecessary($action, $data)
+    {
+
+        if (!isset($data['sendChatAction'])) {
+            return false;
+        }
+
+        if ($data['sendChatAction'] === true) {
```
noplanman: Shouldn't this be `!== false`?

smoqadam: Yeah that's my fault. it should be `!== true`

Issue #869, opened 2018-07-31
Bot API 4.0


noplanman:
https://core.telegram.org/bots/api#july-26-2018

- [ ] Added support for [Telegram Passport](https://telegram.org/blog/passport).
- [ ] Added support for editing the media content of messages: added the method `editMessageMedia` and new types `InputMediaAnimation`, `InputMediaAudio` and `InputMediaDocument`.
- [ ] Added the field `thumb` to the `Audio` object.
- [ ] Added support for attaching custom thumbnails to uploaded files.
- [ ] `tg://` URLs now can be used in inline keyboard `url` buttons and `text_link` message entities.
- [ ] Added the field `animation` to the Message object. For backward compatibility, when this field is set, the `document` field will be also set.
- [ ] Added two new `MessageEntity` types: `cashtag` and `phone_number`.
- [ ] Added support for Foursquare venues: added the new field `foursquare_type` to the objects `Venue`, `InlineQueryResultVenue` and `InputVenueMessageContent`.
- [ ] Added vCard support when sharing contacts: added the field `vcard` to the objects `Contact`, `InlineQueryResultContact` and `InputContactMessageContent`.

PR #870, opened 2018-08-01
Bot API 4.0 - NO Telegram Passport

noplanman:
This PR adds all changes from Bot API 4.0, except the Telegram Passport things, which will be done as a separate PR.

Part of #869


CaliforniaMountainSnake:
When the release)?

PR #871, opened 2018-08-05
Bot API 4.0 & 4.1 - Telegram Passport (without decryption helpers)

noplanman:
Add Telegram Passport.

@php-telegram-bot/developers 
For the errors I've used an interface, like with `InputMedia` and `InputMessageContent`.
Does that make sense, or use an abstract class to extend from?

Also, I'm really not sure how best to test all this (hence WIP) :confused:


noplanman:
@1Ganso Have you tested this code?
If so, would you be able to share your experiences and possibly a short guide on how to set up a testing environment for me to test it?

noplanman:
Merging this without any decryption helpers, which will I've almost finished developing, but will add in a separate PR.

PR #885, opened 2018-09-03
Add missing database columns and methods.

noplanman:
Add missing bits and pieces, mainly database columns and entity methods.


noplanman:
@jacklul Would appreciate your input on this if you have a moment :innocent:

PR #903, opened 2018-10-29
add php 7.3 to travis

andreybolonin:


noplanman:
PHP 7.3 hasn't officially been released yet, so the nightly build should cover that. But as soon as it is released, we can add it :+1:

andreybolonin:
@noplanman `nightly` now is php 7.4-dev ))

PR #689, opened 2017-11-01
issue #323 modify getCommandObject and executeCommand for possibility…

Jonybang:
For issue: Mess with custom command loading (#323)

Was add possibility to use custom namespace for commands. The code is tested on a real project and works well with two parallel bots. Also backward compatibility is observed

PR #919, opened 2019-01-23
Allow to set Custom namespace for commands, also

imami:
Added `setCommandsNamespace()` and `setCommandsPath()` methods to prevent vendors commands and define your own commands outside vendor. before to these changes, you must override vendor's commands like `StartCommand` with the same namespace, wich might cause PHP exception of class redeclaration.

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -61,6 +61,13 @@ class Telegram
     protected $input;
 
     /**
+     * Custom commands namespace
+     *
+     * @var string
+     */
+    protected $commands_namespace = '';
+
+    /**
      * Custom commands paths
      *
      * @var array
@@ -177,6 +184,7 @@ class Telegram
         }
 
         //Add default system commands path
+        $this->commands_namespace = __NAMESPACE__ . '\\Commands\\';
         $this->addCommandsPath(TB_BASE_COMMANDS_PATH . '/SystemCommands');
 
         Request::initialize($this);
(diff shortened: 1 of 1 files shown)


src/Telegram.php
```diff
@@ -649,6 +670,19 @@ public function isDbEnabled()
...
 
+    /**
+     * Set commands path to given array
+     *
+     * @param array $paths  Custom commands paths to set
+     * @return \Longman\TelegramBot\Telegram
+     */
+    public function setCommandsPath(array $paths)
+    {
+        $this->commands_paths = $paths;
```
noplanman: @imami Best to set as an empty array and then use the existing `addCommandsPaths` method to add paths. Just to ensure the reuse of code.

src/Telegram.php
```diff
@@ -277,7 +298,7 @@ public function getCommandObject($command)
         $which[] = 'User';
 
         foreach ($which as $auth) {
-            $command_namespace = __NAMESPACE__ . '\\Commands\\' . $auth . 'Commands\\' . $this->ucfirstUnicode($command) . 'Command';
+            $command_namespace = $this->commands_namespace . $auth . 'Commands\\' . $this->ucfirstUnicode($command) . 'Command';
```
noplanman: I'd vote for having this as `$this->commands_namespace . '\\' . $auth . 'Commands\\'`, to not require the final backslashes when setting the namespace.

noplanman:
@imami Thanks for the PR, but as far as I can see, #689 supersedes this, right?

Or do you have a different scenario that isn't covered by the other PR?

PR #921, opened 2019-01-26
Messages with type of contact now can be handled by ContactCommand.

### LINKED ISSUE #920: Messages with type of `contact` can not be handled
Messages with type of `contact` does not resolve to any of SystemCommands or UserCommands.
so when a user shared his own contact, it can not be handled by this package.

### Required Information
- Operating system:
- PHP version: 7.2
- PHP Telegram Bot version: 0.55.1
- Using MySQL database: no
- Update Method: both `Webhook` and `getUpdates`

### Expected behaviour
It should be processable by a Command (if exists).

### Actual behaviour
This type of messages are not resolved to any command!

### Steps to reproduce
1. send a message with `Keyboard`, and a `KeyboadButton` with `'request_contact' => true`
2. Then you will received a message from user which is type of `contact`

imami:
Fixes #920 
Contact type messages can now be handles if `ContactCommand` defined

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -481,6 +481,7 @@ class Telegram
                 'pinned_message',
                 'invoice',
                 'successful_payment',
+                'contact',
             ], true)
             ) {
                 $command = $this->getCommandFromType($type);


noplanman:
@imami This is now part of version 0.56.0 :ok_hand:

Issue #917, opened 2019-01-07
/cleanup and Integrity constraint violation


Elglobo:
- Operating system: Linux debian Jessie
- PHP version: 7.2
- PHP Telegram Bot version: 0.551
- Using MySQL database: yes
- MySQL version: 10.0.37-MariaDB-0+deb8u1 
- Update Method: Webhook
- Self-signed certificate: yes

### Expected behaviour
Clean old record database

### Actual behaviour
Error on execution:
Warning: PDO::query(): SQLSTATE[23000]: Integrity constraint violation: 1451 Cannot delete or update a parent row: a foreign key constraint fails (`database`.`message`, CONSTRAINT `message_ibfk_5` FOREIGN KEY (`reply_to_chat`, `reply_to_message`) REFERENCES `message` (`chat_id`, `id`))

### Steps to reproduce
Initialize database and after a while error on /cleanup command.

Elglobo:
Any reply or help ?

noplanman:
@Elglobo Sorry for the late reply. I'm travelling at the moment but will check this out when I'm back 👍

jacklul:
My guess we need to remove the records in specific order (replies first), it looks like it tries to remove a message that has replies to

jacklul:
This will be hard to fix as for example you can have very old message that is supposed to be deleted and a reply message that is not supposed to be removed and in this case the you can't remove the old message because of the reply.

We could modify the query for cleaning 'message' so that it cleans newer first (replies for example) but it won't fix the example I mentioned above:
```sql
                'DELETE FROM `%1$s`
                WHERE id IN
                (
                    SELECT id
                    FROM
                    (
                        SELECT ID
                        FROM  `%1$s`
                        WHERE `date` < \'%2$s\'
                        AND `id` NOT IN (
                            SELECT `message_id`
                            FROM `%3$s`
                            WHERE `message_id` = `%1$s`.`id`
                        )
                        AND `id` NOT IN (
                            SELECT `message_id`
                            FROM `%4$s`
                            WHERE `message_id` = `%1$s`.`id`
                        )
                        ORDER BY id DESC
                    ) a
                )
```

Note that this query WAS NOT TESTED.

noplanman:
This has been fixed in #930 and part of the next release :tada:

PR #930, opened 2019-02-28
Possible fix for #917

jacklul:
This should prevent constrain errors, one of side effects observed is that some messages might not get deleted in the first run. #917

Usage of two additional nested subqueries is extremely heavy here though... not sure if doing one select query and then using PHP to select ids to remove wouldn't be a lighter solution here.

PS. telegram_update query was broken because of this missing `AND updated_at < \'%1$s\'` statement for `chat` table - it didn't clean the table until I added that - @noplanman ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 - Botan.io service has been discontinued.
 ### Removed
 ### Fixed
+- Constraint errors in `/cleanup` command.
 ### Security
 
 ## [0.55.1] - 2019-01-06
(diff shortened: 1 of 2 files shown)


noplanman:
Ok, I've looked into this and there is a super simple fix for this:
**Reverse the direction of message deletion!**

Then there are no constraint problems :tada: 

@jacklul I've added a commit to this PR with just that tiny change, please test it to see if it also works for you.

noplanman:
@jacklul Have reverted my changes and left yours, which does the trick :wink: 

Good to merge?

jacklul:
Actually I think the subquery is not required and will only impact performance, works on my end!

noplanman:
@jacklul Doesn't work for me, I get this:
```
Warning: PDO::query(): SQLSTATE[HY000]: General error: 1093 You can't specify target table 'message' for update in FROM clause in /vendor/longman/telegram-bot/src/Commands/AdminCommands/CleanupCommand.php on line 398
```

PR #940, opened 2019-03-22
Autoselect SystemCommand from existing Commands

### LINKED ISSUE #920: Messages with type of `contact` can not be handled

noplanman:
As suggested by @jacklul here: https://github.com/php-telegram-bot/core/issues/920#issuecomment-461529726

Message field types can now automatically choose which command class gets loaded to execute.
If none is defined, GenericmessageCommand is executed by default.
Removed all empty dummy SystemCommands from library.

:exclamation: This can create problems for any project that does not specifically implement SystemCommands that handle all different types, as now GenericmessageCommand gets called for all of them, whereas before they would just be handled silently.
For this reason, SystemCommands must now either be dummied specifically, or GenericmessageCommand must handle the conditions to work correctly.

@jacklul What do you think, does this make sense? Or should we leave the current dummies as they are to preserve backwards compatibility and minimise any possible breakage?

Overrules and closes #921
Fixes #920


jacklul:
Maybe we could add deprecation notice into generic one to notify people about this change?

noplanman:
@jacklul Perfect, I'll do that then for now. At some point they'll be removed completely though.

PR #954, opened 2019-05-29
Logo upload

Maxiride:
Added logo folder with plain and uk flag versions.

Cheers


Maxiride:
Added a commit with the .svg and .ai files.

PR #963, opened 2019-06-07
WIP: Add Database Migrations (And more features as a result)

akalongman:
In this PR added:

- Laravel's Database Query Builder and Migrations [QB Docs](https://laravel.com/docs/5.8/queries)  [Migrations Docs](https://laravel.com/docs/5.8/migrations)
- Laravel's Service Container [CI Docs](https://laravel.com/docs/5.8/container)
- DotEnv Integration (and the possibility to create cacheable config files like Laravel's config)
- Possibility to create CLI Commands like `php telegram blabla`

This brings only one BC change, PHP version should be >=7.1.3, but versions below 7.1 are already deprecated and not supported anymore.

@noplanman @jacklul please take a look, what do you think about the changes

How to test:

Pull this branch, copy `.env.example` to `.env` and write your settings, after run `php telegram migrate`

Two tables should be created, `migrations` for holding versions and `users` table which just copied from Laravel's default migration.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         }
     ],
     "require": {
-        "php": "^5.5|^7.0",
+        "php": "^7.1",
         "ext-pdo": "*",
         "ext-curl": "*",
         "ext-json": "*",
@@ -32,7 +32,7 @@
         "guzzlehttp/guzzle": "^6.3"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8|^5.7|^6.5|^7.5|^8.1",
+        "phpunit/phpunit": "^7.5",
         "squizlabs/php_codesniffer": "^3.4"
     },
     "autoload": {
@@ -58,5 +58,11 @@
         "test-cov-upload": [
             "wget https://scrutinizer-ci.com/ocular.phar && php ocular.phar code-coverage:upload --format=php-clover build/logs/clover.xml"
         ]
+    },
+    "config": {
+        "platform": {
+            "php": "7.1.0"
+        },
+        "process-timeout": 3600
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -23,19 +23,28 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.1.3",
         "ext-pdo": "*",
         "ext-curl": "*",
         "ext-json": "*",
         "ext-mbstring": "*",
         "monolog/monolog": "^1.24",
-        "guzzlehttp/guzzle": "^6.3"
+        "guzzlehttp/guzzle": "^6.3",
+        "illuminate/database": "5.8.*",
+        "illuminate/console": "^5.8",
+        "illuminate/events": "^5.8",
+        "vlucas/phpdotenv": "^3.3",
+        "illuminate/filesystem": "^5.8"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5",
-        "squizlabs/php_codesniffer": "^3.4"
+        "squizlabs/php_codesniffer": "^3.4",
+        "symfony/var-dumper": "^4.3"
     },
     "autoload": {
+        "files": [
+            "src/helpers.php"
+        ],
         "psr-4": {
             "Longman\\TelegramBot\\": "src/"
         }
@@ -61,7 +70,7 @@
     },
     "config": {
         "platform": {
-            "php": "7.1.0"
+            "php": "7.1.8"
         },
         "process-timeout": 3600
     }
--- a/src/Application.php
+++ b/src/Application.php
@@ -1,0 +1,141 @@
+<?php
+declare(strict_types=1);
+
+namespace Longman\TelegramBot;
+
+use Illuminate\Container\Container;
+use Illuminate\Support\Str;
+
+use const DIRECTORY_SEPARATOR;
+
+class Application extends Telegram
+{
+    protected $basePath;
+    protected $appPath;
+    protected $environmentPath;
+    protected $environmentFile = '.env';
+    protected $hasBeenBootstrapped = false;
+
+    public function __construct(string $basePath = null)
+    {
+        if ($basePath) {
+            $this->setBasePath($basePath);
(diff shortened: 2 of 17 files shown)


src/helpers.php
```diff
@@ -0,0 +1,65 @@
+<?php
```
noplanman: Wouldn't these conflict with functions that might be defined by the underlying project already?
Especially as they are necessary for the functioning of the library.

Wouldn't it be better to have a dedicated `Helpers` class here?

src/Bootstrap/LoadEnvironmentVariables.php
```diff
@@ -0,0 +1,39 @@
...
+use Longman\TelegramBot\Application;
+
+class LoadEnvironmentVariables
+{
+    public function bootstrap(Application $app)
+    {
+        try {
+            $this->createDotenv($app)->load();
+        } catch (Exception $e) {
+            $this->writeErrorAndDie($e);
```
noplanman: Does this make sense in respect of having this library integrated as part of an underlying project?
This literally forces the user to have a `.env` file (with that name!) in the current folder.

If I have a custom `whatever.env` that I load in my project, then the PHP Telegram Bot should just pick up those variables.

This bootstrapper should only be run when using the console command I think.

src/Application.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        return $this->basePath . DIRECTORY_SEPARATOR . 'bootstrap' . ($path ? DIRECTORY_SEPARATOR . $path : $path);
+    }
+
+    public function getCachedConfigPath(): string
+    {
+        return $_ENV['APP_CONFIG_CACHE'] ?? $this->bootstrapPath() . '/cache/config.php';
+    }
+
+    public function loadEnvironmentFrom($file): Application
```
noplanman: This method can never be used, as the bootstrapping of the environment file happens in the constructor :thinking: 

How would one define a custom environment file as it is now?

src/Bootstrap/SetupDatabase.php
```diff
@@ -0,0 +1,66 @@
...
+            return $resolver;
+        });
+
+        $app->enableExternalMySql($capsule->getConnection()->getPdo());
+    }
+
+    private function getConfig(): array
+    {
+        return [
+            'driver'         => 'mysql',
```
noplanman: Looking forward to allowing this to be set via environment variables too :innocent: 
(Why now allow all config parameters to be set via env vars? Like charset, collation, engine, etc.)

How easy do you think it will be to shift the whole DB class to use illuminate/database?

composer.json
```diff
@@ -23,19 +23,28 @@
         }
     ],
     "require": {
-        "php": "^5.5|^7.0",
+        "php": "^7.1.3",
```
noplanman: Why exactly 7.1.3?
(I'm asking more out of curiosity)

src/helpers.php
akalongman: We can get rid of these helper functions, or wrap inside our namespace. I think its better than an abstract class with static functions

src/Bootstrap/LoadEnvironmentVariables.php
akalongman: > If I have a custom whatever.env that I load in my project, then the PHP Telegram Bot should just pick up those variables.

src/Application.php
akalongman: I am still thinking about that. How we can give the possibility to users to use some custom env file.

src/Bootstrap/SetupDatabase.php
akalongman: > Looking forward to allowing this to be set via environment variables too

:+1:

> How easy do you think it will be to shift the whole DB class to use illuminate/database?

I think it pretty easy. Let's do that in the second iteration. First, let's make fixes in this integration (think about better .env loading, helper functions, etc.)

composer.json
akalongman: Its Laravel database package's minimum version :)

noplanman: Ok :smiley:

src/Bootstrap/SetupDatabase.php
noplanman: Agreed :ok_hand:

src/Bootstrap/LoadEnvironmentVariables.php
noplanman: Well, that's what the checks in the Telegram constructor are for, right?
If no API key is given, then an exception is thrown.

Same thing happens for the DB.

I think libraries that extend functionality of a project shouldn't require an own .env setup, but instead should define which environment variables are necessary for it to work.

src/helpers.php
noplanman: > or wrap inside our namespace

You mean as functions directly, instead of static methods?

src/Bootstrap/LoadEnvironmentVariables.php
akalongman: @noplanman when user will create an instance of out `Telegram/Application`, our script will check the project's root for `.env` file.

> I think libraries that extend functionality of a project shouldn't require an own .env setup, but instead should define which environment variables are necessary for it to work.

How to achieve this? Do not parse .env and just try to find variables in the $_ENV? Then user should parse .env file himself

src/helpers.php
akalongman: Yes, like \Telegram\app().

Those helpers are for internal use, so in most cases, an user will not use those helpers

noplanman:
What about exporting all DB functionality to a separate package altogether?

I'm just wondering how best to handle custom .env files for an existing project without creating any dependency issues, i.e. main project uses different version of phpdotenv.

hz61p1:
I am against the adoption of this PR, because it is a general-purpose package (CORE), and we must continue to develop it in this direction and not depend on implementation. If you need to integrate it into the framework, then create your own separate package for this, which will provide the necessary integration in a particular ecosystem. The same applies to creating your own framework based on this package.

Issue #994, opened 2019-08-19
Merge !993 break custom system command Newchatmembers. 


panaceya:
### 💥 BC Break Report
Merge !993 break custom system command Newchatmembers. 
Affected to all version started from 0.58.0 to 0.60.0

#### Required Information

|  ?                        |  !
| ---                       | ---
| PHP Telegram Bot version  | > 0.57.0
| PHP version               | 7.3.8
| Update Method             | Webhook
| Self-signed certificate   | no

#### Summary
this expression always false for Newchatmembers
https://github.com/php-telegram-bot/core/blob/06099b100195d49485a49cfeb56977248eb8f557/src/Telegram.php#L471

#### Temporary fix
- downgrade to 0.57.0
or
- add app/Commands/Newchatmembers.php (or another dir)
```php
protected $usage = '/newchatmembers';
```

noplanman:
Hi @panaceya, this is due to the fact that system commands no longer get called directly and instead get handled by `GenericmessageCommand`. The problem was that it was possible to execute any system commands by simply calling that command directly in any chat, e.g. `/newchatmembers`.

This change has been flagged as a deprecation since 0.56.0 ([check changelog](https://github.com/php-telegram-bot/core/blob/master/CHANGELOG.md#0560---2019-04-15)).

You're absolutely right that this BC slipped in a bit too early without mentioning it directly 👍 

I'll correct that and thanks for letting me know!

Here is an example of how previous system commands would now be handled:
https://github.com/php-telegram-bot/support-bot/blob/4deebac7cd2525c0fdb7236150e05620b9747f95/commands/GenericmessageCommand.php#L47

noplanman:
@jacklul Even though we were going towards `GenericmessageCommand` handling all non-user commands, I'm kind of rethinking this 🤔 

Does it maybe make sense to have a `ServicemessageCommand` of type `SystemCommand` that takes care of all non-user messages? It just seems weird to me now, that `GenericmessageCommand` should contain the code to be executed when a new user joins for example. "Generic message" should be just that, a user sent a "generic message" to the bot (be it text or any other type).

What do you think? I'm just trying to make this as user-friendly and logical as possible.

(Of course everyone is welcome to join the discussion 🎉)

jacklul:
I honestly think that change was a good thing since all those extra service fields are still refering `Message` object. 

Maybe instead we could intruduce an extra code in `Genericmessage` that redirect some 'service' message to old system command names?

[...]

PR #999, opened 2019-08-30
Remove obsolete service message system commands

noplanman:
| ?            |  !
|---           | ---
| Type         | bug / improvement
| BC Break     | no
| Fixed issues | #994

#### Summary
Remove all obsolete service message system commands and instead try to execute any user-defined overrides directly in `GenericmessageCommand`.

#### Usage
You need to set `SystemCommand::$execute_deprecated = true;` before handling the request, either in your hook or getUpdates file.

If you have a custom `GenericmessageCommand` implementation, you'll need to add the code snippet from the original `GenericmessageCommand.php` to your own.


jacklul:
I don't think renaming `GenericmessageCommand` to `MessageCommand` is a good idea, might get confusing when people create `/message` commands or something

Assuming we allow users to use their own namespaces (#919) they should be able simply to extend this and `parent::execute()` instead of creating new class and copying code, this also prevents people from removing essential code for handling conversations.

src/Commands/SystemCommands/GenericmessageCommand.php
```diff
@@ -68,9 +68,57 @@ public function execute()
 
         //Fetch conversation command if it exists and execute it
         if ($conversation->exists() && ($command = $conversation->getCommand())) {
-            return $this->telegram->executeCommand($command);
+            return $this->getTelegram()->executeCommand($command);
+        }
+
+        // Try to execute any deprecated system commands.
+        if ($deprecated_system_command_response = $this->executeDeprecatedSystemCommand()) {
```
jacklul: I think `executeNoDb()` should call this as well?

noplanman: Absolutely!

noplanman:
@jacklul I've updated this a bit and moved the code to `SystemCommand`, to make the integration into a custom `GenericmessageCommand` easier.

Also, there's a property `SystemCommand::$execute_deprecated` which needs to be set for this to work. I added this, to not have all the code run for users who don't need this feature.

Any thoughts are highly appreciated! 🙏

I'm also thinking about how best to implement the new command system, to make it logical and easy to use.

PR #1005, opened 2019-10-03
Allow to store in notes unicode, like cyrillic

volandku:
bug / improvement

### DIFF
--- a/src/Conversation.php
+++ b/src/Conversation.php
@@ -219,7 +219,7 @@ class Conversation
     public function update()
     {
         if ($this->exists()) {
-            $fields = ['notes' => json_encode($this->notes)];
+            $fields = ['notes' => json_encode($this->notes,JSON_UNESCAPED_UNICODE)];
             //I can update a conversation whatever the state is
             $where = ['id' => $this->conversation['id']];
             if (ConversationDB::updateConversation($fields, $where)) {
--- a/src/Conversation.php
+++ b/src/Conversation.php
@@ -219,7 +219,7 @@ class Conversation
     public function update()
     {
         if ($this->exists()) {
-            $fields = ['notes' => json_encode($this->notes,JSON_UNESCAPED_UNICODE)];
+            $fields = ['notes' => json_encode($this->notes, JSON_UNESCAPED_UNICODE)];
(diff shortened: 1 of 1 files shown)


noplanman:
@volandku Thanks for your PR!

Does the string not get escaped correctly / what output do you get?
What version of PHP are you using?

`json_decode` *should* restore any saved string correctly.

noplanman:
@volandku Just pinging again here to ask what exactly isn't working.

If something is off, would be great to have this fixed in the next release :muscle:

volandku:
Well, I dont remenber for now ))
Seems that result in cyrillic didn`t store at all or stored in wrong format in DB

noplanman:
@volandku Ok, thanks for getting back to me.

I'll close this off for now, as I can't reproduce this, but please do open it again with some example code if it happens again to you! :smiley:

VSMent:
This is really useful. 
Debug, manual db edit or faster manual search are much better now.

noplanman:
@VSMent Would you mind elaborating?
I assume you're talking about the change in this PR?

VSMent:
@noplanman Well, actually even after development. 
I do not know what are cons of storing data with unescaped unicode. For me understandable cyrillic characters is huge pro.

P.S. What is PR?

noplanman:
PR is short for Pull Request.

@VSMent What I meant to ask is, do you also require the change proposed here to be able to store cyrillic characters to the database? If so, could you provide me with a code example that fails so that I can test this and fix it please?

VSMent:
@noplanman Everything works good. But when using PhpStorm or phpMyAdmin to view database tables I do not understand what is written in those cells.
I get "\u0422\u0430\u043A" instead of "Так" → "Yes" in Ukrainian.
It will be preferable for me (and hope for other Cyrillic language developers) to store unicode unescaped.

noplanman:
Aaah, I see now, thanks for the explanation :+1:

I guess then the only "problem" would be interfaces (for viewing the DB fields) that don't support certain encodings. Not sure if that would actually matter though, as the same would apply in other DB fields like messages or user details with special characters.

@volandku Could you add an entry to the CHANGELOG.md in the "Unreleased" section please?

PR #1008, opened 2019-10-17
Upgrade PHPUnit and PHPCS

akalongman:
This PR upgrades PHPUnit to version 8.x and PHPCS to version 3.5. Also fixed PSR-12 incompatibilities in the codebase. For tests now minimum PHP version is 7.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,21 +11,15 @@ cache:
     - $HOME/.composer/cache
 
 php:
+  - 7.4snapshot
   - 7.3
   - 7.2
-  - 7.1
-  - 7.0
-  - 5.6
-  - 5.5
   - nightly
   - hhvm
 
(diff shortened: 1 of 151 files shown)


.travis.yml
```diff
@@ -22,6 +22,10 @@ php:
 
 matrix:
   allow_failures:
+    - php: 5.5
```
noplanman: Might as well remove these from the tests completely, right?

With this PR, the minimum version of PHP is basically 7.2, as everything lower doesn't get tested any more.
Shall we step up the move to PHP 7.2 as a minimum requirement for the core library?

akalongman: I think we should upgrade minimum PHP version to 7.2 for 1.x release only. Now lets use for development/testing only

PR #1013, opened 2019-10-31
Fix Tidelift link

phated:
@noplanman I was noticing that your Tidelift marketing task wasn't being marked as complete because the value didn't match the one we were looking for. I've corrected it here, then we should be able to unlock the new enterprise language.


noplanman:
Hi @phated, thanks for letting me know!

Just recently I've started moving to an organisation [.github repo](https://github.com/php-telegram-bot/.github), so have fixed it there.

I'll remove the overriding funding.yml in this repo in the next release :+1:

Issue #1014, opened 2019-10-31
Exception should not be swallowed when creating command instance


CyberKoo:
### 🎉 Feature Request
using exception chaining
#### Summary

https://github.com/php-telegram-bot/core/blob/46646884f5c05f1161f10605a0705aa24df3e583/src/Telegram.php#L252

this line may be changed to 
`throw new TelegramException('Error getting commands from path: ' . $path, $e);`

attach the original exception can be provide more information to the developer.

noplanman:
@CyberKoo Thanks for the suggestion, this is a good idea!

Feel free to make a PR if you like :+1:

PR #1030, opened 2019-12-12
chain the exception throwed when creating command instance

CyberKoo:
| ?            |  !
|---           | ---
| Type         |  improvement
| BC Break     | no
| Fixed issues | #1014


noplanman:
Thanks @CyberKoo

Issue #527, opened 2017-06-07
Restrict BOT Access


asif-rahman:
### Required Information

- Operating system: Windows
- PHP version: 5.12
- PHP Telegram Bot version: Latest
- Using MySQL database: yes
- MySQL version:
- Update Method: Webhook
- Self-signed certificate: yes
- RAW update (if available):

### Extra details
I want to restrict my BOT to certain groups only. I can do that by checking the chat id against the group id. But the problem is I have too many commands and I don't want to add access code to each of the PHP scripts.
I want to implement the code on a single page, like the hook. But how can I do that in the hook or in any other single page?

jacklul:
This should put you in right direction:

```php
$POST = file_get_contents("php://input");
$POST_DATA = json_decode($POST, true);

$user_id = null;
if (isset($POST_DATA['callback_query']['from']['id'])) {
    $user_id = $POST_DATA['callback_query']['from']['id'];
} elseif (isset($POST_DATA['inline_query']['from']['id'])) {
    $user_id = $POST_DATA['inline_query']['from']['id'];
} elseif (isset($POST_DATA['message']['from']['id'])) {
    $user_id = $POST_DATA['message']['from']['id'];
}

$allowed = [];

if (!in_array($user_id, $allowed)) {
exit;
}
```

The code should be before handling the webhook in hook script.

noplanman:
I like this very much, would be cool to implement this into the manager 😇

TheXron:
> This should put you in right direction:
> 
> ```
> $POST = file_get_contents("php://input");
> $POST_DATA = json_decode($POST, true);
> 
> $user_id = null;
> if (isset($POST_DATA['message']['chat']['id'])) {
>     $user_id = $POST_DATA['message']['chat']['id'];
> }
> 
> $allowed = [];
> 
> if (!in_array($user_id, $allowed)) {
> exit;
> }
> ```
> 
> You will have to modify it a bit to consider other types of updates.
> 
> The code should be before handling the webhook in hook script.

Hello bro. How can I set condition if the array is false? I mean I want to send message to user if they are trying to access my bot.

jacklul:
> Hello bro. How can I set condition if the array is false? I mean I want to send message to user if they are trying to access my bot.

That code is very old, the preferred way of achieving this would be extending Telegram class, example approach:

```php
<?php

class TelegramBot extends Telegram
{
    public function executeCommand(string $command): ServerResponse
    {
[...]

PR #1045, opened 2020-01-21
Feature/33 update filter

VRciF:
The goal of this PR is to provide a method to filter update messages using a custom callback which solves #527.

It is mentioned in https://github.com/php-telegram-bot/telegram-bot-manager/issues/33#issue-235095581 that this might be implemented in the telegram bot manager, I don't use the bot manager currently and only need the core running on my raspberry pi.

| ?            |  !
|---           | ---
| Type         | feature
| BC Break     | no
| Fixed issues | #527 

#### Summary

A new method `setUpdateFilter($callback)` is introduced.If the $callback returns false, the executed of processUpdate() ends with a ServerResponse where isOk() returns false.
The given test case demonstrates the usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 ### Notes
 - [:ledger: View file changes][Unreleased]
 ### Added
+- New method `setUpdateFilter($callback)` used to filter `processUpdate(Update $update)` calls. If `$callback` evaluates to false the update is not processed and a `ServerResponse()` is returned where `->isOk()` is false.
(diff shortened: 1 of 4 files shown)


noplanman:
@VRciF Sorry for this epically late reply...

I like this! Could you please rebase and add a note to the changelog as well?

@php-telegram-bot/developers What do you think about this?

VRciF:
@noplanman No worry about the late reply. We're all too busy too often. I've rebased and modified the changelog. Please have a look.

noplanman:
@VRciF Thanks so much for this!

I've updated it a bit and added the ability to define a custom reason text to be written to the debug log.

Issue #1050, opened 2020-02-07
null properties because of strtolower


enescakir:
#### Required Information

|  ?                        |  !
| ---                       | ---
| Operating system          | Centos 7
| PHP Telegram Bot version  | v1.1.1
| PHP version               | 7.3.14
| MySQL version             | none
| Update Method             | Webhook
| Self-signed certificate   |  no
| RAW update (if available) | -

#### Summary

I am trying to get `chat_id` in a custom command.
```
$chatID = $this->getMessage()->getChat()->getId();
```
But it returns `null` for `chat_id` all the time if language is set to Turkish.

I tracked issue and `strtolower('getId')` returns `getId` in Turkish instead of `getid` in `Entity` class.

`mb_strtolower('getId')` workings corrects and returns `getid`.

#### Current behaviour
It returns null `chat_id`.

#### Expected behaviour
It should return `chat_id`.

PR #1051, opened 2020-02-07
change lowercase function to mb_strtolower instead of strtolower

enescakir:
| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | no
| Fixed issues | #1050

#### Summary
I am trying to get `chat_id` in a custom command.
```
$chatID = $this->getMessage()->getChat()->getId();
```
But it returns `null` for `chat_id` all the time if language is set to Turkish.

I tracked issue and `strtolower('getId')` returns `getId` in Turkish instead of `getid` in `Entity` class.

`mb_strtolower('getId')` workings corrects and returns `getid`.


enescakir:
PTAL @noplanman 👀

noplanman:
Thanks @enescakir :tada:

PR #1074, opened 2020-04-13
Use class constants instead of strings for 'generic' and 'genericmessage'

1int:
## Summary
This is a trivial change so I decided not to make an issue for it.

This PR replaces occurrences of 'generic' and 'genericmessage' with class constants to improve code clarity and prevent typos. Having class constants with proper comments also helps understand the expected behaviour.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 ### Notes
 - [:ledger: View file changes][Unreleased]
 ### Added
+- Replaced 'generic' and 'genericmessage' strings with Telegram::GENERIC_COMMAND and Telegram::GENERIC_MESSAGE_COMMAND constants (@1int)
 ### Changed
 ### Deprecated
 ### Removed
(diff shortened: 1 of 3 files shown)


noplanman:
Thanks, hadn't thought of that!

I do think however, that it makes more sense to use `GENERIC_COMMAND` instead of `FALLBACK_COMMAND`, because fallback could mean anything.

Also, please add a short entry in the changelog in the `Unreleased - Added` section, thanks.

1int:
Yeah I was doubting between `GENERIC_COMMAND` and `FALLBACK_COMMAND` myself :) 
I chose FALLBACK_COMMAND because I think it reflects the point of that command better — if nothing else suits, fall back to this. 

Ok, I'll make these changes today. Cheers!

1int:
Updated

1int:
@noplanman hey, I also thought that now that there're constants it would possibly be good to rename 'genericmessage' to 'genericMessage' camel-case? Then you could have more consistent naming with **GenericMessageCommand** instead of **GenericmessageCommand**.

I can make another PR if you think it makes sense.

noplanman:
The whole command system is quite complex and needs a rewrite in general, so I think it makes sense to leave it as is for the moment, but thanks so much for the suggestion!

Issue #1088, opened 2020-05-02
Log request data sended to Telegram API


asafov:
#### Summary

I think it will be good idea to log request data sended to telegram server in debug log.
Maybe at dedicated "Verbose" log.

noplanman:
Do you mean to output the `$data` values in the error log along with the HTTP response?

asafov:
Yes

noplanman:
I like this idea, thanks for bringing it up. I'll prepare a PR for you to check out 👌

noplanman:
@asafov What do you think, regarding the PR?

asafov:
Looks great. Fully as i wanted.

jacklul:
> Looks great. Fully as i wanted. Maybe add setter and getter? What do you think?

Since it's a static class it doesn't make sens, you can simply:

```php
TelegramLog::$log_request_data = true;
```

and change this at any time on run-time.

asafov:
OK. Merge.

asafov:
I create PR for log response data too.

PR #1089, opened 2020-05-02
Allow logging request data in the debug log

noplanman:
| ?            |  !
|---           | ---
| Type         | feature
| BC Break     | no
| Fixed issues | #1088 

#### Summary

Add a new `TelegramLog::$always_log_request_and_response` variable, to output request data to the debug log before posting the request itself to Telegram.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 ### Added
 - Replaced 'generic' and 'genericmessage' strings with Telegram::GENERIC_COMMAND and Telegram::GENERIC_MESSAGE_COMMAND constants (@1int)
 - Bot API 4.8 (Extra Poll and Dice features).
+- `TelegramLog::$log_request_data` parameter to output the request data to the debug log.
 ### Changed
 ### Deprecated
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 ### Added
 - Replaced 'generic' and 'genericmessage' strings with Telegram::GENERIC_COMMAND and Telegram::GENERIC_MESSAGE_COMMAND constants (@1int)
 - Bot API 4.8 (Extra Poll and Dice features).
-- `TelegramLog::$log_request_data` parameter to output the request data to the debug log.
+- `TelegramLog::$always_log_request_and_response` parameter to force output of the request and response data to the debug log, also for successful requests
 ### Changed
 ### Deprecated
 ### Removed
--- a/doc/01-utils.md
+++ b/doc/01-utils.md
@@ -40,10 +40,10 @@ Telegram API changes continuously and it often happens that the database schema 
 If you store the raw data you can import all updates on the newest table schema by simply using [this script](../utils/importFromLog.php).
 Remember to always backup first!!
 
-### Request data
-If you'd like to also log the request data to your debug log, you can set the appropriate variable:
+### Always log request and response data
+If you's like to always log the request and response data to the debug log, also for successful requests, you can set the appropriate variable:
 ```php
-\Longman\TelegramBot\TelegramLog::$log_request_data = true;
+\Longman\TelegramBot\TelegramLog::$always_log_request_and_response = true;
 ```
 
 [PSR-3]: https://www.php-fig.org/psr/psr-3
--- a/doc/01-utils.md
+++ b/doc/01-utils.md
@@ -40,6 +40,11 @@ Telegram API changes continuously and it often happens that the database schema 
 If you store the raw data you can import all updates on the newest table schema by simply using [this script](../utils/importFromLog.php).
 Remember to always backup first!!
 
+### Request data
+If you'd like to also log the request data to your debug log, you can set the appropriate variable:
+```php
+\Longman\TelegramBot\TelegramLog::$log_request_data = true;
+```
 
 [PSR-3]: https://www.php-fig.org/psr/psr-3
 [PSR-3-providers]: https://packagist.org/providers/psr/log-implementation
--- a/src/Request.php
+++ b/src/Request.php
@@ -486,10 +486,6 @@ class Request
         $request_params['debug'] = TelegramLog::getDebugLogTempStream();
 
         try {
-            if (TelegramLog::$log_request_data) {
-                TelegramLog::debug('Request Data: ' . print_r($data, true));
-            }
-
             $response = self::$client->post(
                 '/bot' . self::$telegram->getApiKey() . '/' . $action,
                 $request_params
@@ -504,6 +500,9 @@ class Request
             $result = $e->getResponse() ? (string) $e->getResponse()->getBody() : '';
         } finally {
             //Logging verbose debug output
+            if (TelegramLog::$log_request_data) {
+                TelegramLog::debug('Request Data:' . PHP_EOL . print_r($data, true));
+            }
             TelegramLog::endDebugLogTempStream('Verbose HTTP Request output:' . PHP_EOL . '%s' . PHP_EOL);
         }
 
--- a/src/Request.php
+++ b/src/Request.php
@@ -486,6 +486,10 @@ class Request
         $request_params['debug'] = TelegramLog::getDebugLogTempStream();
 
         try {
+            if (TelegramLog::$log_request_data) {
+                TelegramLog::debug('Request Data: ' . print_r($data, true));
+            }
+
             $response = self::$client->post(
                 '/bot' . self::$telegram->getApiKey() . '/' . $action,
                 $request_params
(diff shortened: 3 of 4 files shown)


asafov:
> LGTM but what about logging what `$action` was requested?

Action? You mean method? It's present in request URL.

jacklul:
> > LGTM but what about logging what `$action` was requested?
> 
> Action? You mean method? It's present in request URL.

I don't think it's in `$data` though!

asafov:
> > > LGTM but what about logging what `$action` was requested?
> > 
> > 
> > Action? You mean method? It's present in request URL.
> 
> I don't think it's in `$data` though!

What you mean under "Action"? Example please.

noplanman:
@jacklul As @asafov mentioned, the action / method is in the post URL.

My initial thought was to keep the `$data` output to show only the parameters that have actually been sent as the data body.

You think it should be added to the `$data` debug output?

asafov:
This information for example helps me to debug MarkdownV2 chars.
Without it i will not be able to fix it.

jacklul:
> @jacklul As @asafov mentioned, the action / method is in the post URL.
> 
> My initial thought was to keep the `$data` output to show only the parameters that have actually been sent as the data body.
> 
> You think it should be added to the `$data` debug output?

Actually I was worried that sometimes the record with URL can not be just before the data one when bot is heavily used, then I remember you could use Monolog's UidProcessor to know which records were in the same request...

noplanman:
@jacklul I now understand what you mean!! I've moved the request data output to "just before" the actual response, so if a separate process gets in-between those, all bets are lost anyway 😅

asafov:
So what about log response data too?

noplanman:
@asafov The HTTP response gets logged just after the (now) request data and the response body itself gets logged in the `update` log, read here:
https://github.com/php-telegram-bot/core/blob/78024d7bc55cebf3987247b465a1009b5186060a/doc/01-utils.md

Or do you mean something more than that?

asafov:
I mean this situation for example:

```
[2020-05-03 03:37:05] production.DEBUG: Request Data: Array
(
    [chat_id] => 837010404
    [text] => %5Byandexnavi%3A%2F%2Fbuild_route_on_map%3Flat_to%3D55.822313%26lon_to%3D37.377214%5D%28%D0%9E%D1%82%D0%BA%D1%80%D1%8B%D1%82%D1%8C+%D0%AF%D0%BD%D0%B4%D0%B5%D0%BA%D1%81.%D0%9D%D0%B0%D0%B2%D0%B8%D0%B3%D0%B0%D1%82%D0%BE%D1%80%29
    [parse_mode] => MarkdownV2
)
  
[2020-05-03 03:37:05] production.DEBUG: Response Body: {"ok":false,"error_code":400,"description":"Bad Request: can't parse entities: Character '.' is reserved and must be escaped with the preceding '\\'"}
  
[2020-05-03 03:37:05] production.DEBUG: Verbose HTTP Request output:
* About to connect() to api.telegram.org port 443 (#1489)
*   Trying 149.154.167.220...
* Connected to api.telegram.org (149.154.167.220) port 443 (#1489)
*   CAfile: /etc/pki/tls/certs/ca-bundle.crt
  CApath: none
* SSL connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate:
* 	subject: CN=api.telegram.org,OU=Domain Control Validated
* 	start date: Mar 24 13:48:17 2020 GMT
* 	expire date: May 23 16:17:38 2022 GMT
* 	common name: api.telegram.org
* 	issuer: CN=Go Daddy Secure Certificate Authority - G2,OU=http://certs.godaddy.com/repository/,O="GoDaddy.com, Inc.",L=Scottsdale,ST=Arizona,C=US
> POST /bot916473975:AAFPmfGbuxVr15DrgIfJzLWp7L6Ww4uqyI0/sendMessage HTTP/1.1
User-Agent: GuzzleHttp/6.5.3 curl/7.29.0 PHP/7.2.27
Content-Type: application/x-www-form-urlencoded
Host: api.telegram.org
Content-Length: 382

* upload completely sent off: 382 out of 382 bytes
< HTTP/1.1 400 Bad Request
< Server: nginx/1.16.1
< Date: Sun, 03 May 2020 00:37:05 GMT
< Content-Type: application/json
< Content-Length: 150
< Connection: keep-alive
< Strict-Transport-Security: max-age=31536000; includeSubDomains; preload
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Content-Length,Content-Type,Date,Server,Connection
< 
* Connection #1489 to host api.telegram.org left intact
```

asafov:
If query isn't successfully no any update log will be writed.

noplanman:
Ah yes, of course! Thought that was in the cURL response 🤦
(but only the reason is stated there, like `HTTP/1.1 400 Bad Request`)

asafov:
Now you understand what i want? But my PR is incorrect too - it's log all responses, not only errors. Can you fix it? And i think that for errors not need param for log or not, coz it's real error.

noplanman:
The way it is now, the debug log contains all requests, successful and failed.

I think you're right in that only errors should be logged, for `de-bugging` purposes.
([Which is confirmed on Wikipedia](https://en.wikipedia.org/wiki/Syslog#Severity_level))

So it's probably best to change this behaviour altogether.
@jacklul What do you think? This would obviously be a BC change in the sense that logged output would be different from now on, but I must admit that the noise makes the debug log a difficult to use. Or introduce another variable to allow ALL logging? I guess that's what `TRACE` would be for, but [we don't have that available](https://www.php-fig.org/psr/psr-3/#5-psrlogloglevel).

jacklul:
@asafov I've removed bot token you accidentaly posted with the log.

@noplanman Any chance we could filter out bot tokens from the output ?
I think simple switch variable would be okay, or option to pass separate Logger for http requests debug.

noplanman:
Thanks for removing the key @jacklul.
@asafov Since GitHub remembers all edits and your API key can still be viewed, please reset it with `@BotFather`.

I like the idea of removing the API key from the debug output, but I'm just wondering of a use-case where it might actually help to see it to debug a problem 🤔 
Should be easy to do with regex.

Regarding the separate logger, you're suggesting another logger instance on top of the two already set when initialising the logger?

asafov:
@jacklul thank you very much.
@noplanman done, thank you.

jacklul:
@noplanman Actually I just found out you can remove edit revisions too

noplanman:
@jacklul Cool, thanks for the info, didn't know that!
Ah nice, a PR maybe for the API token removal?

I've updated this PR to include a variable `TelegramLog::$always_log_request_and_response` to define if request and response data should be output to the debug log, even when it's a successful request. Failed requests always get logged.

@asafov Could you have a quick look at this before I merge please?

jacklul:
> @jacklul Cool, thanks for the info, didn't know that!
> Ah nice, a PR maybe for the API token removal?

I'll PR when I verify it doesn't remove anything else except token, so far I can see it messes up the dates

noplanman:
Right. It would be nice if we had access to `Telegram::getApiKey` to literally just remove the API key, but we don't 😕 

It might need more specific parameters, like `'~/bot(\d+)\:[\w\-]+/~'` (untested)

jacklul:
Actually I'm thinking if we could use second parameter of `debug()` function to post the array data to?

noplanman:
@jacklul You mean like this
```php
TelegramLog::debug("Request data:\n{data}", ['data' => print_r($data, true)]);
TelegramLog::debug("Response data:\n{data}", ['data' => $result]);
```
instead of this?
```php
TelegramLog::debug('Request data:' . PHP_EOL . print_r($data, true));
TelegramLog::debug('Response data:' . PHP_EOL . $result);
```

Not sure that helps with code readability though.

jacklul:
> @jacklul You mean like this
> ...

No, second argument for methods in PsrLog is context array.

```php
TelegramLog::debug('Request data', $data);
TelegramLog::debug('Response data:' . PHP_EOL . $result);
TelegramLog::debug('Response data' , ['result' => $result]);
```
```
[2020-05-04 12:52:39] test.DEBUG: Request data {"offset":0,"limit":null,"timeout":null} []
[2020-05-04 12:52:39] test.DEBUG: Response data: 
{"ok":true,"result":[]} [] []
[2020-05-04 12:52:39] test.DEBUG: Response data {"result":"{\"ok\":true,\"result\":[]}"} []
```

> 
> Not sure that helps with code readability though.

It actually makes things even worse 😇
The only advantage is that (depending on logger) this can be used to further process the array/data.

I'd say leave it as it is now, if anyone requests it we could implement a switch for this in the future.

noplanman:
I see, your code takes advantage of Monologs way of converting parameters into strings, in this case [converting the `$data` array to JSON](https://github.com/Seldaek/monolog/blob/bec314a9c14ce8a40650cf13923f5941ef1bfe0a/src/Monolog/Formatter/LineFormatter.php#L155). [Apix does the same](https://github.com/apix/log/blob/d94aaefcc6e900535317b25e213fa59a0a555215/src/LogFormatter.php#L49).

This only works for frameworks with such behaviour, which (as far as I know) is not part of the PSR-3 spec. It's kind of "misusing" the context parameters.

We'll leave it as it is in that case.
Ready to merge then? 😃

Issue #1093, opened 2020-05-03
MarkdownV2 support?


asafov:
At telegram api docs present parse_mode = MarkdownV2.
I tried to send some data and find, that needed new escape for support this format.

https://core.telegram.org/bots/api#markdownv2-style

```php
    public static function escapeMarkdownV2($text)
    {
        $markdown = [
            '#',
            '*',
            '_',
            '=',
            '.',
            '[',
            ']',
            '(',
            ')',
            // ... rest of markdown entities
        ];
        //'_‘, ’*‘, , ’~‘, ’`‘, ’>‘, ’#‘, ’+‘, ’-‘, ’=‘, ’|‘, ’{‘, ’}‘, ’.‘, ’!‘
        $replacements = [
            '\#',
            '\*',
            '\_',
            '\\=',
            '\.',
            '\[',
            '\]',
            '\(',
            '\)',
            // ... rest of corresponding escaped markdown
        ];

        return str_replace($markdown, $replacements, $text);
    }
```

Something like this, as i think.

noplanman:
@asafov Have created a PR for you to review 👌

devope:
Are you planning to support auto escaping?
May be i create a PR for it? 

I think it would be really convenient to pass second argument for `markdownV2` for escaping

genesiscz:
@devope THIS!

PR #1094, opened 2020-05-03
Add static method `Entity::parseMarkdownV2` to handle MarkdownV2.

noplanman:
| ?            |  !
|---           | ---
| Type         | improvement
| BC Break     | yes
| Fixed issues | #1093 

#### Summary
- Add new static method `Entity::parseMarkdownV2` to handle MarkdownV2.
- Change `Entity::parseMarkdown` method to be static, thus not requiring an explicit object.


asafov:
I not sure, that my function is full. Need to check it.

noplanman:
I have taken all characters from the official docs for this PR, so should be complete.

PR #1095, opened 2020-05-04
Remove bot token from debug stream

jacklul:
This will remove bot token from debug http log streams, specifically from the URL path, I don't think there is any other place the token shows up?

So this:
```
> POST /bot123456789:AAHc2FVEadqyhkTYcUuG1RiYR4Hfgpwc/sendMessage HTTP/1.1
```
becomes this:
```
> POST /botBOT_TOKEN_REMOVED/sendMessage HTTP/1.1
```

Can be disabled on demand with `TelegramLog::$remove_bot_token = false`.
Having it enabled as default will make it safe for the people to post these logs for troubleshooting.

Regex is literally modified one from `Telegram.php:172`.
My original idea was to cache the token in the class to not run regex for each entry but people might be running some weird setups where they handle multiple bots in single script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 - Bot API 4.8 (Extra Poll and Dice features).
 - Allow custom MySQL port to be defined for tests.
 - New static method `Entity::escapeMarkdownV2` for MarkdownV2.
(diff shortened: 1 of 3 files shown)


noplanman:
Thanks for the PR 👍 

I agree with not caching the token, as that would also just add an extra variable somewhere.
Also, as this code is only executed when a request fails, it should be fine regarding performance.

Issue #1099, opened 2020-05-20
Idea: Ability to pause / resume conversation


asafov:
I want to realise feature for pause / resume conversation.
Use case:
For example i have bot with catalog of goods, listed by inline controls.
User see needed item and click "Add to basket". Bot ask quantity of items and save it to storage and pause conversation, for user can continue list catalog from last position.
When user want to add another item - conversation resume and etc.
And the finish line user view him basket and finish order (stopped) or cancel it.

What do you think, guys?

noplanman:
I think that's a great idea and a nice addition 🎉 

Feel free to create a PR and some example code so we can test it out.

asafov:
OK. Will start develop at this night.

PR #1102, opened 2020-05-26
Pause / resume functionality

asafov:
Pause / resume functionality for #1099

### DIFF
--- a/src/Conversation.php
+++ b/src/Conversation.php
@@ -69,17 +69,18 @@ class Conversation
      * @param int    $user_id
      * @param int    $chat_id
      * @param string $command
+     * @param bool $paused
      *
      * @throws TelegramException
      */
-    public function __construct($user_id, $chat_id, $command = null)
+    public function __construct($user_id, $chat_id, $command = null, $paused = false)
     {
         $this->user_id = $user_id;
         $this->chat_id = $chat_id;
         $this->command = $command;
 
         //Try to load an existing conversation if possible
-        if (!$this->load() && $command !== null) {
+        if (!$this->load($paused) && $command !== null) {
             //A new conversation start
             $this->start();
         }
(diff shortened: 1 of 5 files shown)


asafov:
Call example:
`$this->conversation = new Conversation($chatId, $chatId, $this->getName(), true);`

noplanman:
Thanks for the PR @asafov 

I've noticed that you've worked on the `master` branch and pointed this PR to `master` as well.

As all work is done on the `develop` branch and only finished releases go into `master`, please rebase your commits (onto the `develop` branch) and change the PR to point to `develop`.

:pray:

Looking forward to give this a test afterwards :smiley:

asafov:
> Thanks for the PR @asafov
> 
> I've noticed that you've worked on the `master` branch and pointed this PR to `master` as well.
> 
> As all work is done on the `develop` branch and only finished releases go into `master`, please rebase your commits (onto the `develop` branch) and change the PR to point to `develop`.
> 
> 
> 
> Looking forward to give this a test afterwards

Done.

noplanman:
As far as i recall how conversations are set up, there can only be 1 conversation per user per command, so there can never be multiple conversations happening for the same command.
For this reason, the state of a conversation can only be a single state, never multiple ones.

With that in mind, a paused conversation is the same like an active one, but with a `paused` status. So when requesting a conversation, there is no need to know in which state the conversation is in, as there will only be 1 to return anyway, and then the state can be read from that.

So looking at the PR, it seems that most of the extra code for the `paused` status isn't necessary and can be accomplished much more easily.

If I've missed something or have it wrong, please explain :pray:

asafov:
So when user start conversation just get active or paused like this?
`            $sql = '
              SELECT *
              FROM `' . TB_CONVERSATION . '`
              WHERE (`status` = :status OR `status` = :pstatus)
                AND `chat_id` = :chat_id
                AND `user_id` = :user_id
            ';

            if ($limit !== null) {
                $sql .= ' LIMIT :limit';
            }

            $sth = self::$pdo->prepare($sql);

            $sth->bindValue(':status', 'active');
            $sth->bindValue(':pstatus', 'paused');`

noplanman:
@asafov Looking better indeed 👍 

Just a few things:
1. When pausing and resuming, the conversation doesn't need to be cleared, as that unnecessarily re-fetches all data from the DB.
2. When updating the status, the status of the `$conversation` array need to be updated too (as it wouldn't be reloaded after changing point 1.)

Looking at the conversation feature in general, I think it would be great to export it into a separate little package and redesign it a bit. For instance, having the `$conversation` variable seems a bit weird and it probably makes sense to extract all data directly into the `Conversation` class, to not have that second layer.

What do you think?

We could then also extend the class to have `getStatus` or even `isActive`/`isPaused` methods.

asafov:
All fixes it applied. Can anybody check it?

asafov:
Guys?

noplanman:
@asafov I've updated the code a tiny bit, could you have a look please?

If all is ok with your tests, we can get this merged 💪

noplanman:
After some deliberation, it looks like the conversations don't need a `pause` state per se.

What should be updated though, is to allow parallel commands to be active, 1 per command.
At the moment any active conversation gets cancelled when a new one is started.

Also it would make sense to extract the conversations into a separate package entirely.

PR #1118, opened 2020-06-23
Custom namespace fixes again...

jacklul:
`getCommandObject()` will now return a valid object when using custom namespace (instead of NULL because `$filepath` is not provided)

Not sure if this is secure as it skips check for admin commands access this way.

Previously: https://github.com/php-telegram-bot/core/pull/1115

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -295,6 +295,10 @@ class Telegram
      */
     public function getCommandObject($command, $filepath = null)
     {
+        if (isset($this->commands_objects[$command])) {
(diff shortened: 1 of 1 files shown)


noplanman:
I think it should be ok, as that list is generated taking the admin status into account. ([Admin check just before loading the commands objects](https://github.com/php-telegram-bot/core/blob/3134bdeeed59aa836aee2562615706d46b469568/src/Telegram.php#L534))

So if the user isn't an admin, the command won't be in that list anyway.

PR #1130, opened 2020-09-28
feat: upgrade guzzlehttp/guzzle to 7.x

KristobalJunta:
| ?            |  !
|---           | ---
| Type         | feature
| BC Break     | yes
| Fixed issues | #1127 

#### Summary

Upgrade Guzzle HTTP version to 7.x. This should be considered breaking change, as Guzzle dropped PHP 5 support and minimum compatible PHP version is now 7.2.

This change allows to use php-telegram-bot with Laravel 8.x apps.

Issue #1127, opened 2020-09-24
Upgrade to Guzzle 7.x


KristobalJunta:
#### Summary

Laravel 8 got recently released and ships with Guzzle version 7.x out of the box. For now, the package can't be used in a new Laravel 8 project as the current stable release (0.63.1) requires Guzzle ^6.3.

PR #1133, opened 2020-10-03
feat: upgrade guzzlehttp/guzzle to 7.x

KristobalJunta:
| ?            |  !
|---           | ---
| Type         | feature
| BC Break     | no
| Fixed issues | #1127 

#### Summary

Add support for Guzzle HTTP version 7.x. This change allows to use php-telegram-bot with Laravel 8.x apps. Not breaking backwards compatibility here, because project works both with 6.x and 7.x Guzzle versions.

An improvement over #1130, which was backwards incompatible.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "ext-json": "*",
         "ext-mbstring": "*",
         "psr/log": "^1.1",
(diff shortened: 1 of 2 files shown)


noplanman:
Thanks @KristobalJunta 👍

PR #1143, opened 2020-10-27
Code quality fixes

akalongman:
Add code quality improvements

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -25,75 +25,163 @@ use Longman\TelegramBot\Exception\TelegramException;
 /**
  * Class Request
  *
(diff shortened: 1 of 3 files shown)


noplanman:
@akalongman Regarding the reformatting of the docblock, to me it looks quite messy with the new lines.
Is there a specific reason to reformat like this?

For what it is, i personally don't mind that it exceeds the recommended line length, as it's not code per se.

akalongman:
@noplanman on large screens it's pretty readable, but maybe you are right and we should wrap long lines, even in doc-blocks :) Let's leave as-is for now and change if someone will complain about that

noplanman:
I was suggesting to **NOT** wrap the docblock lines, to keep the class header clean.
I really doubt anybody reads that, as it's the same text from the Bot API docs, which is where users will check about parameters etc anyway.

Issue #1062, opened 2020-03-27
broken markup on long messages


kur1977:
Hi there ))
**Request::sendMessage**

```php
        do {
            //Chop off and send the first message
            $data['text'] = mb_substr($text, 0, 4096);
            $response     = self::send('sendMessage', $data);

            //Prepare the next message
            $text = mb_substr($text, 4096);
        } while (mb_strlen($text, 'UTF-8') > 0);
```

This breaks markup on long texts. 
I remove this and found that texst that are longer than 4096 symbols are successfully delivers to the chat.

So, may be this part of code is a bug or something?

jacklul:
That code cuts the text but doesn't account for any markdown or HTML markup, if it cuts text while a tag is still open it will error out. In this case you should handle it yourself with your own helper function

noplanman:
According to [the docs](https://core.telegram.org/bots/api#sendmessage) the message (without any markup) can be no longer than 4096 characters.

@kur1977 Out of curiosity, how long is the message you're successfully sending when you remove the snippet?

kur1977:
@noplanman 
After removing this part of code I succesfully sended the marked-up messages that more than twice bigger than 4096 charecters.

noplanman:
@kur1977 I've updated the method in #1163 to allow for splitting or not, returning all message responses.

With that I'll close this off and let users manage their own message sending.

Issue #923, opened 2019-02-01
Return array of responses on sendMessage


Abduvokhid:
Please, change `Request::sendMessage` to return the array of responses in case if the text is longer than 4096 symbols. Now, this method is returning the response of last sent message.

![image](https://user-images.githubusercontent.com/16443812/52114760-b8a9d600-262e-11e9-8e99-7b13314838f9.png)

jacklul:
Ideally it would be good to keep this backwards compatible by adding responses into separated `ServerResponse` field and making root `ok` field return `True` when all messages where sent and `False` when at least one failed.

noplanman:
I agree that this should change.

I can see some different ways of solving this:
- return an array or `ServerResponse` (breaks backwards compatibility)
- return a single custom response object, ok if all messages sent successfully, response objects nested in root response (what @jacklul mentioned above)
- add an extra parameter by reference, which contains all response objects, e.g. `sendMessage($data, $responses);` (doesn't break backwards compatibility, but not sure if it's the prettiest)

@jacklul So you're saying to add an extra custom field to the `ServerResponse` class?

Abduvokhid:
Or you can add an array of responses into last|first response of `sendMessage`. This array will be set if the text is divided into parts and sent. In this case, the developer has to check if this array is set or not.

vontend:
I have just get problem with this case. A work around is split your message send overloop and return the last one

PR #1163, opened 2020-12-23
Return all sendMessage response objects in referenced array variable

noplanman:
| ?            |  !
|---           | ---
| Type         | improvement
| BC Break     | no
| Fixed issues | #923 #1062

#### Summary

When sending long messages with `Request::sendMessage(...)` all response objects are now returned in an extra parameter which is passed by reference:
```
Request::sendMessage($data, $extras);
// `$extras['responses']` lists all `ServerRequest` response objects for split message sending
```

@Abduvokhid I know it's been a long time, but what do you think of this solution?
@jacklul Would appreciate any feedback on this, thanks 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 ### Notes
 - [:ledger: View file changes][Unreleased]
 ### Added
-- Extra parameter for `Request::sendMessage()` to return all response objects for split messages.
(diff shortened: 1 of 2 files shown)


Abduvokhid:
@noplanman looks good.

But as you commented markdown can be broken.

noplanman:
@Abduvokhid Yes, not sure how to fix that to make it bullet-proof.
I've added the ability to pass options to the `sendMessage` method, which gives the user a bit more freedom.

Best would be to handle special cases manually though.

PR #1184, opened 2021-03-05
Fixed keyboard constructur without any parameters

hutattedonmyarm:
| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | yes
| Fixed issues | 1183

#### Summary

Fixed Keyboard and InlineKeyboard constructor without parameters

### DIFF
--- a/src/Entities/Keyboard.php
+++ b/src/Entities/Keyboard.php
@@ -110,7 +110,7 @@ class Keyboard extends Entity
         }
 
         // If $args was empty, $data still contains `false`
-        return $data ?? [];
+        return $data ?: [];
     }
 
     /**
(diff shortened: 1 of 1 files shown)


noplanman:
@hutattedonmyarm Perfection!! I was literally checking the code and noticed you've pushed a corrected version, which is exactly how I would have fixed it 💪 

Thanks so much for this!

src/Entities/Keyboard.php
```diff
@@ -109,7 +109,8 @@ protected function createFromParams(): array
             $data[$keyboard_type] = $new_keyboard;
         }
 
-        return $data;
+        // If $args was empty, $data still contains `false`
+        return $data ?? [];
```
noplanman, line 113: Ah, this should be `?:` instead of `??`.

hutattedonmyarm: You're right, I only tested it in PHP's interactive shell, and due to a typo it worked there! Fixed now

noplanman (approved):

Issue #1186, opened 2021-03-09
Bot API 5.1


noplanman:
https://core.telegram.org/bots/api#march-9-2021

:exclamation: WARNING! :exclamation:
After one of the upcoming Bot API updates, user identifiers will become bigger than 2^31 - 1 and it will be no longer possible to store them in a signed 32-bit integer type. User identifiers will have up to 52 significant bits, so a 64-bit integer or double-precision float type would still be safe for storing them. Please make sure that your code can correctly handle such user identifiers.

**Added two new update types**

- [x] Added updates about member status changes in chats, represented by the entity `ChatMemberUpdated` and the fields `my_chat_member` and `chat_member` in the `Update` entity.

**Improved Invite Links**

- [x] Added the entity `ChatInviteLink`, representing an invite link to a chat.
- [x] Added the method `createChatInviteLink`, which can be used to create new invite links in addition to the primary invite link.
- [x] Added the method `editChatInviteLink`, which can be used to edit non-primary invite links created by the bot.
- [x] Added the method `revokeChatInviteLink`, which can be used to revoke invite links created by the bot.

**Voice Chat Info**

- [x] Added the type `VoiceChatStarted` and the field `voice_chat_started` to the entity `Message`.
- [x] Added the type `VoiceChatEnded` and the field `voice_chat_ended` to the entity `Message`.
- [x] Added the type `VoiceChatParticipantsInvited` and the field `voice_chat_participants_invited` to the entity `Message`.
- [x] Added the new administrator privilege `can_manage_voice_chats` to the entity `ChatMember` and parameter `can_manage_voice_chats` to the method `promoteChatMember`.

**And More**

- [x] Added the type `MessageAutoDeleteTimerChanged` and the field `message_auto_delete_timer_changed` to the entity `Message`.
- [x] Added the new administrator privilege `can_manage_chat` to the entity `ChatMember` and parameter `can_manage_chat` to the method `promoteChatMember`.

noplanman:
@massadm I'll finish implementing everything here, thank you so much for getting this started 🙏

PR #1192, opened 2021-03-21
Schema migration

massadm:
| ?            |  !
|---           | ---
| Type         | feature
| BC Break     | no
| Fixed issues | #1186 

#### Summary

Schema migration

### DIFF
--- a/utils/db-schema-update/0.71.0-unreleased.sql
+++ b/utils/db-schema-update/0.71.0-unreleased.sql
@@ -1,0 +1,48 @@
+# Invite Links
+CREATE TABLE IF NOT EXISTS `chat_invite_link` (
+    `id` BIGINT UNSIGNED COMMENT 'Unique identifier for this entry'
+        PRIMARY KEY,
+    `invite_link` VARCHAR(2083) CHARACTER SET 'ascii' COLLATE 'ascii_general_ci' NOT NULL COMMENT 'The invite link. If the link was created by another chat administrator, then the second part of the link will be replaced with “…”',
+    # IE7 has a 2083 character limit for HTTP GET operations: http://support.microsoft.com/kb/208427
+    `creator_id` BIGINT NOT NULL COMMENT 'Creator of the link',
+    `is_primary` BOOLEAN NOT NULL COMMENT 'True, if the link is primary',
+    `is_revoked` BOOLEAN NOT NULL COMMENT 'True, if the link is revoked',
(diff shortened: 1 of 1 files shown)


massadm:
Please review and let me know if changes can be moved to the `structure.sql`.

utils/db-schema-update/0.71.0-unreleased.sql
```diff
@@ -0,0 +1,48 @@
+# Invite Links
+CREATE TABLE IF NOT EXISTS `chat_invite_link` (
+    `id` BIGINT UNSIGNED COMMENT 'Unique identifier for this entry'
+        PRIMARY KEY,
+    `invite_link` VARCHAR(2083) CHARACTER SET 'ascii' COLLATE 'ascii_general_ci' NOT NULL COMMENT 'The invite link. If the link was created by another chat administrator, then the second part of the link will be replaced with “…”',
```
noplanman, line 5: We should be safe with much less here, I really don't think an invite link will have that many characters.

Also, why the collation change for this field?

massadm: > We should be safe with much less here, I really don't think an invite link will have that many characters.

_VARCHAR values are stored as a one-byte or two-byte length prefix plus data. A VARCHAR column uses one length byte if values require no more than 255 bytes, two length bytes if values may require more than 255 bytes._

So, `VARCHAR(255)` or `VARCHAR(2083)` is real options. I leave it up to you.

> Also, why the collation change for this field?

I am assuming here that the URI will be stored in transport form which is 7-bit coded character set, and in addition, non-ASCII internationalized domain names intended for resolution via the DNS will be IDNA encoded as well.

noplanman: Ok, thanks for the explanation. I think 255 and staying with `utf8mb4` is more than enough.

What advantage do you see in using `ascii` vs. the `utf8mb4`?

massadm: > What advantage do you see in using ascii vs. the utf8mb4?

UUIDs or URIs are clearly ASCII so CHARACTER SET 'ascii' is suitable domain for storing data like this.

noplanman:
@massadm Thanks for your contribution, you can go ahead and add the new tables and fields to the `structure.sql` file 🙏 

Remember that the tables also need to be populated in `DB.php`, but I'm happy to do that part, which can be a bit confusing with the codebase as it is now 😁

PR #1199, opened 2021-03-30
Bot API 5.1

noplanman:
| ?            |  !
|---           | ---
| Type         | feature / improvement
| BC Break     | no
| Fixed issues | #1186

#### Summary

Add all features for Bot API version 5.1 (March 2021)


noplanman:
@massadm, would you mind checking this PR over so I can then merge it?

You'll notice that I've removed the `chat_invite_link` table and save the JSON text directly. This is how other fields are saved as well. For sure the whole DB structure needs refactoring and the objects could be saved differently, but for now I think it's best to keep it simple.

massadm (commented):
OK

Issue #1201, opened 2021-03-30
Add `allowed_updates` parameter for getUpdates method


noplanman:
#### Summary

The [`getUpdates`](https://core.telegram.org/bots/api#getupdates) method must allow for the `allowed_updates` parameter to be passed.

This is vital to allow the new `chat_member` update type to be retrieved.

Looking at the current [`handleGetUpdates`](https://github.com/php-telegram-bot/core/blob/9ca6158f9fe6cf5aa06f9c46377b3c1164bdf444/src/Telegram.php#L400) method, best solution would be to add an extra array parameter to pass the allowed update types.

massadm:
This is the correct way to implement `update_filter`. Check https://github.com/php-telegram-bot/core/blob/9ca6158f9fe6cf5aa06f9c46377b3c1164bdf444/src/Telegram.php#L528

noplanman:
@massadm, so what you're saying, is to define all known types specifically [here](https://github.com/php-telegram-bot/core/blob/9ca6158f9fe6cf5aa06f9c46377b3c1164bdf444/src/Telegram.php#L444) and the user can then filter by using the provided callback function?

massadm:
`update_filter` is `null` by default so no one update will be missed. If some updates is not welcome the right policy is not to receive them at all. Decision can be made after each iteration with help of `setUpdateFilter`.

noplanman:
The problem here is, that for the `chat_member` update type, the value **must** be defined in the `allowed_updates` field. As it is now, it will never get that update, so the `update_filter` also never gets it.

Check #1202

massadm:
Btw, found an error in the following piece of code:
```php
    /**
     * Update filter
     * Filter updates
     *
     * @var callback
     */
    protected $update_filter;
```
`@var` must be `callable` here, not "callback".

massadm:
> The problem here is, that for the `chat_member` update type, the value **must** be defined in the `allowed_updates` field. As it is now, it will never get that update, so the `update_filter` also never gets it.

I see. Let it remain so: an empty list to get all types of updates except for `chat_member`, until the user changes it with `update_filter`.

massadm:
> Check #1202

IMO, update types better to implement as constants and make use of `update_filter` instead of an additional parameter. Bot can use `setUpdateFilter` method to change behavior at any time.

noplanman:
Hi @massadm, thanks for your input, I've reflected your thoughts in the PR.

[...]

PR #1202, opened 2021-03-30
Allow defining update types when handling updates via `getUpdates` method

noplanman:
| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | no
| Fixed issues | #1201 

#### Summary

Add an extra parameter to `Telegram::handleGetUpdates` to define allowed types. If no update types are passed, allow ALL update types, to make sure that special ones like `chat_member` (which MUST be defined to receive) also get fetched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Exclamation symbols (:exclamation:) note something of importance e.g. breaking c
 - Bot API 5.1 (ChatMember Update types, Improved Invite Links, Voice Chat). (@massadm, @noplanman)
 ### Changed
 ### Deprecated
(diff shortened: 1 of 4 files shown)


massadm (commented):
I agree with changes.

One remark. `allowed_updates` is optional and can be empty list to receive all update types except `chat_member` and this should be the default behavior.

noplanman:
@massadm I understand what you mean. The main reason I did it like this, is that only updates that are "known" by the library come through, to prevent any breakage in the future when new types get defined.

I agree with you though, that the default behaviour should be what Telegram sends without specifying anything, will change this 👍

PR #1207, opened 2021-04-03
PSR4 command classes include

alligator77:
| ?            |  !
|---           | ---
| Type         | bug + feature
| BC Break     | yes
| Fixed issues | 1205

#### Summary
Add support to include commands classes via PSR4, and old method (pathes) works too.
Fix regex for search 'namespace' 
Add tests.

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -15,6 +15,7 @@ defined('TB_BASE_PATH') || define('TB_BASE_PATH', __DIR__);
 defined('TB_BASE_COMMANDS_PATH') || define('TB_BASE_COMMANDS_PATH', TB_BASE_PATH . '/Commands');
 
 use Exception;
+use InvalidArgumentException;
 use Longman\TelegramBot\Commands\AdminCommand;
 use Longman\TelegramBot\Commands\Command;
 use Longman\TelegramBot\Commands\SystemCommand;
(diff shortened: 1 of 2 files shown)


noplanman:
Thanks @alligator77, will check this out in the coming days!

noplanman:
Hi @alligator77, thanks for the PR!

A few things:

The command name and auth should be inferred from the command class, as they are defined there already. You'll need to check that the passed class correctly extends `Longman\TelegramBot\Commands\Command` and from there you can create a new dummy object to get the extra info:
```php
$class_name = 'What\\Ever\\MyCommand'; // as an example

if (!is_a($class_name, Command::class)) {
    // return, as not the correct type...
}

$command_object = new $class_name($this); // "$this" here is the instance of Telegram

// Get the name from the command object itself.
$name = $command_object->getName();
```

You can figure out the `$auth` by checking the [`Command::is...Command()` methods](https://github.com/php-telegram-bot/core/blob/9ca6158f9fe6cf5aa06f9c46377b3c1164bdf444/src/Commands/Command.php#L336), like `$command_object->isSystemCommand()`.

```php
$auth = null;
switch (true) {
    case $command_object->isSystemCommand():
        $auth = 'System';
        break;
    case $command_object->isAdminCommand():
        $auth = 'Admin';
        break;
    case $command_object->isUserCommand():
        $auth = 'User';
        break;
}

if ($auth) {
    $this->commandsClasses[$auth][$command] = $className;
}
```

This will cut down on superfluous data required to be entered by the user, as it's there already.

Hope you know what I mean 😃

alligator77:
Fair enough. I was not aware of the implementation of these methods.

noplanman:
Thanks @alligator77 👍

PR #1233, opened 2021-07-03
0.2 compatibility

daniel-lucio:
I understand that $callback_query has all the information, but $query_data has the specific answer,
many examples still expect the second argument, and since it is a URL query strict formatted one, it needs to be converted somehow to an array

| ?            |  !
|---           | ---
| Type         | bug / feature / improvement
| BC Break     | yes / no
| Fixed issues |


noplanman:
Thanks for your input @daniel-lucio.

The callback data is simply a string, so can contain anything really.
It's up to the developer to decide how to manage that data. Therefore, using `parse_str` to force a specific format doesn't make sense.

Hope that makes sense 😊

PR #1252, opened 2021-10-29
feat: add resetCommandsPaths

wright-tw:
Solve the problem that the default command cannot be removed

| ?            |  !
|---           | ---
| Type         | feature 
| BC Break     | yes
| Fixed issues | Solve the problem that the default command cannot be removed

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -866,7 +866,7 @@ class Telegram
         if (!is_dir($path)) {
             TelegramLog::error('reset commands path "' . $path . '" does not exist.');
         } elseif (!in_array($path, $this->commands_paths, true)) {
-            $this->commands_paths = $path;
+            $this->commands_paths = [$path];
         }
 
         return $this;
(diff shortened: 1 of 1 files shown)


noplanman:
Hi @wright-tw, thanks for your contribution and sorry for the long wait!

What would make more sense in my opinion, is to have a `setCommandsPath(string $path)` and `setCommandsPaths(array $paths)` methods, which literally just overwrite the variable, using the existing methods to add the new ones.

Does that make sense to you?

wright-tw:
Hi @noplanman, thanks for your comment.

My purpose of adding a new reset method is
Don't want to use the old built-in commands in the suite

Examples of built-in commands: 
```
src/Commands/AdminCommands/ChatsCommand.php
src/Commands/AdminCommands/CleanupCommand.php
```

noplanman:
@wright-tw Yes, that's how I understood it 👍 

It makes more sense to add this feature more streamlined and to use the already existing methods.
I'll merge your PR and then adjust it.

Thanks for your contribution!

PR #1283, opened 2022-02-11
Create Seen Uz bot

Bek124:
| ?            |  !
|---           | ---
| Type         | bug / feature / improvement
| BC Break     | yes / no
| Fixed issues |


jacklul:
I think you targeted wrong repository.
PS. Do not embed API key into the code as it is considered a secret, now that you posted it publicly anyone can use it - you should reset it!

PR #1291, opened 2022-03-08
Add PHP 8.1 build, exclude composer.lock from git

osavchenko:
| ?            |  !
|---           | ---
| Type         | improvement
| BC Break     | no

#### Summary

Add PHP 8.1 build, exclude `composer.lock` from git


TiiFuchs:
@osavchenko Can you explain why you want to remove the composer.lock file from the repository?

osavchenko:
From my experience `composer.lock` is redundant in the external libraries, it makes libraries more flexible. For example, it allows supporting more PHP versions without bumping any dependencies (the current lock doesn't allow PHP 8.1 out of the box).

Also, without a lock file, it will be possible to make builds on the latest dependencies version

TiiFuchs:
Hmm... When you add a library as a dependency the specific versions in the composer.lock file are ignored.
See https://getcomposer.org/doc/02-libraries.md#lock-file

Since travis should use the oldest resolvable dependencies for php-telegram-bot on its tests I think it doesn't matter for the tests either. 

But I'm okay with removing the composer.lock file... For development and running tests we should also test against newest dependencies.

@noplanman You okay with this too?

TiiFuchs (approved):
I approve this message.

Issue #1295, opened 2022-03-18
change $api_base_uri in order to use Request class with my own server url(tdlib)


arashnz:
### 🐞 Bug Report

When I use 
\Longman\TelegramBot\Request::setCustomBotApiUri(
    $api_base_uri          = 'http://myownserver:8081', // Default: https://api.telegram.org
);
and then use  \Longman\TelegramBot\Request::sendAudio it still use https://api.telegram.org as base api url.
but when I change 'private static $api_base_uri = 'http://myownserver:8081';' in request.php manually it works!

#### Required Information

| Operating system          | Centos 8
| PHP Telegram Bot version  | latest
| PHP version               | 7.4
| MySQL version             | none
| Update Method             | Webhook
| Self-signed certificate   | not sure/use it on my own server

#### Expected behaviour

When using \Longman\TelegramBot\Request::setCustomBotApiUri(
    $api_base_uri          = 'http://myownserver:8081', // Default: https://api.telegram.org
);
for all of the usage of Request:: in that page it should use http://myownserver:8081 as base api telegram url.

TiiFuchs:
As you can see the setCustomBotApiUrl method does exactly the same as setting the variable:
https://github.com/php-telegram-bot/core/blob/72528b72cbc576ebb86652a2541838817c0a0bec/src/Request.php#L324-L330

But the Request::initialize method copies the base_uri from the static property into a Client object.
See https://github.com/php-telegram-bot/core/blob/3cf50e76c3a4809e476b4cd274a3b22783b78882/src/Request.php#L302-L306

This initialize() method gets called in the constructor of the Telegram class...

So you need to call Request::setCustomBotApiUri BEFORE you instantiate a new Telegram class.

```
\Longman\TelegramBot\Request::setCustomBotApiUri('http://myownserver:8081/');
$bot = new Longman\TelegramBot\Telegram($token, $username);
```

For the future I made a note and will have a look into that, so we may loose that requirement or streamline it a bit more.

arashnz:
Yes. problem solved.
Thank you.

PR #1296, opened 2022-03-18
Make sure setCustomBotApiUri has an effect if Telegram object was already instantiated.

TiiFuchs:
Should fix problems like #1295

### DIFF
--- a/src/Request.php
+++ b/src/Request.php
@@ -327,6 +327,10 @@ class Request
         if ($api_base_download_uri !== '') {
             self::$api_base_download_uri = $api_base_download_uri;
         }
+
+        if (self::$telegram) {
+            self::initialize(self::$telegram);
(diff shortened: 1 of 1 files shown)


TiiFuchs:
@noplanman What do you think?

TiiFuchs:
We have a few difficulties with finding the right way here, because you also can set a Guzzle Client yourself with `Request::setClient()` and this should always take precedence over custom settings, like the base_uri we're setting in `setCustomBotApiUri()`.

For now I added a check that looks, if there is already a client and throws an `TelegramException`, so we get an Exception if we set the custom Bot API URI after instantiating the Telegram class with an understandable error that helps solve the problem directly.

If you use your own Guzzle Client, you need to set the base_uri accordingly yourself!

PR #1298, opened 2022-03-20
Docker integration

TiiFuchs:
I added a first Docker integration.

Just a php container with database. No webserver for now. Maybe we can add one later for development or when we need to test webhooks.

I changed the database constants in the phpunit.xml.dist accordingly and also changed the composer scripts so it makes use of the container. Eventually we need to revise this again.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^9.5",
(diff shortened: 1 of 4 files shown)


TiiFuchs:
## How to use

- Run `docker-compose up -d` to build and start up the container
- Run `composer run-script test` to run tests

That's it. Should be pretty straight-forward.

PR #1318, opened 2022-04-17
Bot API 6.0

TiiFuchs:


src/Entities/Factory.php
```diff
@@ -6,9 +6,11 @@ abstract class Factory
 {
     abstract public static function make(array $data, string $bot_username): Entity;
 
-    public static function resolveEntityClass(string $class, array $property, string $bot_username = ''): Entity
+    public static function resolveEntityClass(string $class, mixed $property, string $bot_username = ''): Entity
     {
-        if (is_subclass_of($class, Factory::class)) {
+        if (is_a($property, $class)) {
+            return $property;
+        } elseif (is_subclass_of($class, Factory::class)) {
```
TiiFuchs, line 13: I added this check, so we can use subEntities() on classes that don't come from Telegram like in KeyboardButton. So if the user decides to set an object property with an array like this `setWebApp(['url' => 'https://example.com'])` it will still gets parsed into a WebAppInfo object if the user calls `getWebApp()` later on...

noplanman: Fancy!

Tiny change to have cleaner returns:
```suggestion
        }
        
        if (is_subclass_of($class, Factory::class)) {
```

src/Request.php
```diff
@@ -297,7 +312,7 @@ class Request
     /**
      * Initialize
      *
-     * @param Telegram $telegram
+     * @param  Telegram  $telegram
```
noplanman, line 315: We need to agree on some standard here, these things just bloat a PR so much :see_no_evil: 

Looks like most changes in this file are cosmetic, some even unpretty.

noplanman:
This is looking good, nice work :+1: 

Last thing missing is the DB migrations to rename all the `voice_chat` -> `video_chat` fields. And to add the new WebApp fields.

@TiiFuchs You wanna give it a go?

src/Request.php
TiiFuchs: Errg yes, that was not on purpose... Seems like i hit the auto format in PhpStorm here... I'll clean that up.

TiiFuchs:
> This is looking good, nice work 👍
> 
> Last thing missing is the DB migrations to rename all the `voice_chat` -> `video_chat` fields. And to add the new WebApp fields.
> 
> @TiiFuchs You wanna give it a go?

I always forget about the database >.<"

I can try this evening, yeah. :)

TiiFuchs:
> @TiiFuchs You wanna give it a go?

@noplanman Done. Did I miss anything?

TiiFuchs:
I think we need to check `createResultObject `and `createResultObjects`

TiiFuchs:
> I think we need to check `createResultObject `and `createResultObjects`

Yup there was stuff missing. I added that.
Should be finished now.

PR #1335, opened 2022-06-27
Make public to edit commands_objects

jyxjjj:
When do not use ```runCommands``` or ```processUpdate```,  ```$command_object``` cannot be updated so cannot run ```executeCommand```:
It appears like this:
- create a new Command
- call it by executeCommand
- this command's __construct can be run
- this commans's execute cannot be run

Due to:
https://github.com/php-telegram-bot/core/blob/0.77.1/src/Telegram.php#L610
https://github.com/php-telegram-bot/core/blob/0.77.1/src/Telegram.php#L661

| ?            |  !
|---           | ---
| Type         |  improvement
| BC Break     | no
| Fixed issues |

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -99,7 +99,7 @@ class Telegram
      *
      * @var array
      */
-    protected $commands_objects = [];
+    public $commands_objects = [];
 
     /**
      * Current Update object


jyxjjj:
Temp fix (in the project that this core sdk be used):
```
(fn()=>$this->command_objects = $this->getCommandList())->call($telegram);
```

jyxjjj:
Fix way 2:
```
<?php

namespace App\Common;

use GuzzleHttp\Client;
use Longman\TelegramBot\Entities\ServerResponse;
use Longman\TelegramBot\Exception\TelegramException;
use Longman\TelegramBot\Telegram as BaseTelegram;

class Telegram extends BaseTelegram
{
    private static function getClient(): Client
    {
        return new Client([
            'base_uri' => env('TELEGRAM_API_BASE_URI'),
            'proxy' => env('TELEGRAM_PROXY'),
            'timeout' => 60,
        ]);
    }

    /**
     * @throws TelegramException
     */
    public static function getTelegram(): Telegram
    {
        $telegram = new self(env('TELEGRAM_BOT_TOKEN'), env('TELEGRAM_BOT_USERNAME'));
        Request::setClient(self::getClient());
        return $telegram;
    }

    public function executeCommand(string $command): ServerResponse
    {
        $this->commands_objects = $this->getCommandsList();
        return parent::executeCommand($command);
    }
}

```

jyxjjj:
But this two are not a good way

jyxjjj:
fix way 3:
```
$property = new ReflectionProperty(Telegram::class, 'commands_objects');
$property->setValue($telegram, $telegram->getCommandsList());
$telegram->executeCommand($message->getCommand());
```

jyxjjj:
Sorry, i think it should be a issue, i don't know how to fix it due to Choice Difficulty and Obsessive-Compulsive Disorder

TiiFuchs:
I’m not understanding what problem you’re trying to solve here.

jyxjjj:
> I’m not understanding what problem you’re trying to solve here.

If i do not use ```runCommands``` or ```processUpdate```
I will not go to run something like https://github.com/php-telegram-bot/core/blob/0.77.1/src/Telegram.php#L610
So i cannot run command with only ```executeCommand```
If i did, the ```Command Class```'s ```__construct``` will be ran, but the ```execute``` will not.
That means if i only call ```executeCommand```,
I will not got things like https://github.com/php-telegram-bot/core/blob/0.77.1/src/Telegram.php#L1218 did.

jyxjjj:
> > I’m not understanding what problem you’re trying to solve here.
> 
> If i do not use `runCommands` or `processUpdate` I will not go to run something like https://github.com/php-telegram-bot/core/blob/0.77.1/src/Telegram.php#L610 So i cannot run command with only `executeCommand` If i did, the `Command Class`'s `__construct` will be ran, but the `execute` will not. That means if i only call `executeCommand`, I will not got things like https://github.com/php-telegram-bot/core/blob/0.77.1/src/Telegram.php#L1218 did.

Then, It is necessary to go back to how i got this matter. Until the root cause is that I want to get rid of the existing commands in the vendor.

PR #1356, opened 2022-08-29
Keyboard addRow array support

jyxjjj:
| ?            |  !
|---           | ---
| Type         | improvement
| BC Break     | no
| Fixed issues | null

#### Summary

make addRow both support button args and array of buttons.

## I haven't test it.
I am sure it will not break this file but i don't know if there is any file else used this function like override or something, if so, it will break.

Please help me finish this.

### DIFF
--- a/src/Entities/Keyboard.php
+++ b/src/Entities/Keyboard.php
@@ -120,12 +120,17 @@ class Keyboard extends Entity
      *
      * @return Keyboard
      */
-    public function addRow(): Keyboard
+    public function addRow($button, ...$buttons): Keyboard
     {
-        if (($new_row = $this->parseRow(func_get_args())) !== null) {
-            $this->{$this->getKeyboardType()}[] = $new_row;
+        if (is_array($button)) {
+            if (($new_row = $this->parseRow($button)) !== null) {
+                $this->{$this->getKeyboardType()}[] = $new_row;
+            }
+        } else {
+            if (($new_row = $this->parseRow(func_get_args())) !== null) {
+                $this->{$this->getKeyboardType()}[] = $new_row;
+            }
         }
-
         return $this;
     }
 


jyxjjj:
Hummm, it seems checker broken when starting, instead of code check failed, i think it should be re-run.

TiiFuchs:
Looks to me like you're completely ignoring `$buttons` here... 
You may want to revise this.

jyxjjj:
No, if the first param is array, all button should be supplied in the first param, if not, func_get_args will got all the buttons. So the second param is actually useless.

TiiFuchs:
Oh yeah, I've misread the func_get_args there...

The problem is though, that `is_array($button)` will alway be true, because a button is also just an array, an array of buttons is still an array. So the whole else part ist kind of never used in both scenarios...

jyxjjj:
Sorry but, i saw this
```
InlineKeyboardButton extends KeyboardButton
KeyboardButton extends Entity
Entity implements \JsonSerializable
```
My test code is:
```
class A implements \JsonSerializable
{
    public function jsonSerialize(): array
    {
        return [];
    }
}
echo is_array(new A()) ? '1' : '0';
```
Then it returns 0,
> My mind: how could it be an array?
-------
And i saw your code just now:
```
Request::sendMessage([
    'chat_id'      => 12345,
    'text'         => 'Test',
    'reply_markup' => [
        'inline_keyboard' => [
            [ // Row 1
              [ // Button 1 in 1st row
                'text'          => 'A',
                'callback_data' => 'A'
              ],
            ],
        ]
    ]
]);
```
Okay, then it is an array now
I never thought this way...
But, i think it still works,
Let me think, i will update later.

TiiFuchs:
Yeah, that's kind of a "problem" with the library for now...
A lot of stuff can be replaced with pure arrays...

We currently don't have a strong object coupling with keyboards...
The fluent-keyboard package tries to make this a bit sturdier.

jyxjjj:
Ok, i tried many ways, include this:
```
    public function addRow($buttons): Keyboard
    {
        if (is_array($buttons)) {
            $is_instance = true;
            foreach ($buttons as $button) {
                if (!($button instanceof KeyBoardButton)) {
                    $is_instance = false;
                }
            }
            if ($is_instance && ($new_row = $this->parseRow($buttons)) !== null) {
                $this->{$this->getKeyboardType()}[] = $new_row;
            }
        }
        if (($new_row = $this->parseRow(func_get_args())) !== null) {
            $this->{$this->getKeyboardType()}[] = $new_row;
        }
        return $this;
    }
```
And this:
```
    public function addRow(array $buttons): Keyboard
    {
        foreach ($buttons as $button) {
            if (!($button instanceof KeyboardButton)) {
                return $this;
            }
        }
        if (($new_row = $this->parseRow($buttons)) !== null) {
            $this->{$this->getKeyboardType()}[] = $new_row;
        }

        return $this;
    }
```
This:
```
    public function addRow($buttons): Keyboard
    {
        if (!is_array($buttons)) $buttons = [$buttons];
        foreach ($buttons as $button) {
            if (!($button instanceof KeyboardButton)) {
                if (($new_row = $this->parseRow(func_get_args())) !== null) {
                    $this->{$this->getKeyboardType()}[] = $new_row;
                }
                return $this;
            }
        }
        if (($new_row = $this->parseRow($buttons)) !== null) {
            $this->{$this->getKeyboardType()}[] = $new_row;
        }

        return $this;
    }
```
It is really hard to compatible with all situations,
Can we just make a BC break?


I submit this due to laravel's model's where function,
it can be used as:
```
->where('column', $data)
// or
->where('column', '=', $data)
// or
->where([
'column' => $data,
])
```

Any more opinion or advice?

PR #1365, opened 2022-10-07
Enable using commands containing underscores

### LINKED ISSUE #1364: Unable to define commands containing underscore
#### Required Information

|  ?                        |  !
| ---                       | ---
| Operating system          | n/a
| PHP Telegram Bot version  | 0.79.0
| PHP version               | 7.4.32
| MySQL version             | none
| Update Method             | Webhook
| Self-signed certificate   | no
| RAW update (if available) | n/a

#### Summary

The [specs says this](https://core.telegram.org/bots#commands):

> Commands can use latin letters, numbers and underscores

But it's not possible to define a command that contains underscores.

For example, let's assume that we'd like to define the command `/foo_bar`.

If we define it in a file named `FooBarCommand.php`, at [this line](https://github.com/php-telegram-bot/core/blob/0.79.0/src/Telegram.php#L286) we have that `$command` has a value of `Foobar`, and [`$command_name`](https://github.com/php-telegram-bot/core/blob/0.79.0/src/Telegram.php#L287) will be `foobar`.

If we define it in a file named `Foo_BarCommand.php`, at [this line](https://github.com/php-telegram-bot/core/blob/0.79.0/src/Telegram.php#L286) we have that `$command` has a value of `FooBar`, and [`$command_name`](https://github.com/php-telegram-bot/core/blob/0.79.0/src/Telegram.php#L287) will still be `foobar`.

#### Current behaviour

It's impossible to define a command containing underscores.

#### How to reproduce

Create a command like this one

```php
<?php

class FooBarCommand extends UserCommand
{
    protected $name = 'foo_bar';
}
```

It won't be executed with `/foo_bar`.

#### Expected behaviour

Invoking `/foo_bar` should execute a command defined in a file named `FooBarCommand`.

mlocati:
| ?            |  !
|---           | ---
| Type         | improvement
| BC Break     | yes
| Fixed issues | #1364

#### Summary

As explained in #1364, it's currently impossible to define commands containing underscores.

This PR enables it.

For example:

- a `FoobarCommand` class defines the `/foobar` command
- a `FooBarCommand` class defines the `/foo_bar` command (this is a BC break)

PS: closes #1364

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -283,10 +283,9 @@ class Telegram
 
                 foreach ($files as $file) {
                     //Remove "Command.php" from filename
-                    $command      = $this->sanitizeCommand(substr($file->getFilename(), 0, -11));
-                    $command_name = mb_strtolower($command);
+                    $command      = $this->classNameToCommandName(substr($file->getFilename(), 0, -4));
 
-                    if (array_key_exists($command_name, $commands)) {
+                    if (array_key_exists($command, $commands)) {
                         continue;
                     }
 
@@ -294,7 +293,7 @@ class Telegram
 
                     $command_obj = $this->getCommandObject($command, $file->getPathname());
                     if ($command_obj instanceof Command) {
-                        $commands[$command_name] = $command_obj;
+                        $commands[$command] = $command_obj;
                     }
                 }
             } catch (Exception $e) {
@@ -334,7 +333,7 @@ class Telegram
             return null;
         }
 
-        $command_class = $command_namespace . '\\' . $this->ucFirstUnicode($command) . 'Command';
+        $command_class = $command_namespace . '\\' . $this->commandNameToClassName($command);
 
         if (class_exists($command_class)) {
             return $command_class;
@@ -678,7 +677,7 @@ class Telegram
     }
 
     /**
-     * Sanitize Command
+     * @deprecated
      *
      * @param string $command
      *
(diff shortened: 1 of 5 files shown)


noplanman:
Thanks @mlocati, sorry for the huge delay in replying to your PR.

@TiiFuchs and @jacklul, what do you think? I'm wondering if the BC break will be a problem :thinking:

src/Telegram.php
```diff
@@ -1285,4 +1284,56 @@ public function getUpdateFilter(): ?callable
...
+    /**
+     * Converts the name of a class into the name of a command.
+     *
+     * @param string $class For example FooBarCommand
+     *
+     * @return string for example foo_bar. In case of errors, returns an empty string
+     */
+    protected function classNameToCommandName(string $class): string
+    {
+        if (!preg_match('/^(.+)Command$/', $class, $matches)) {
```
TiiFuchs, line 1297: I see no advantage for preg_match over substr comparisons. I guess the latter would be more efficient. May not matter, but regex seems overkill here.

src/Telegram.php
```diff
@@ -1285,4 +1284,56 @@ public function getUpdateFilter(): ?callable
...
+            return '';
```
TiiFuchs, line 1298: Returning an empty string if no command is found, is not good style. Should return null instead.

src/Telegram.php
```diff
@@ -1285,4 +1284,56 @@ public function getUpdateFilter(): ?callable
...
+                $chunks[] = $currentUpperCaseLetter;
+            }
+            $temp = substr($temp, $upperCaseLetterOffset + strlen($upperCaseLetter));
+            $currentUpperCaseLetter = $upperCaseLetter;
+        }
+        $lastChunk = $currentUpperCaseLetter . $temp;
+        if ($lastChunk !== '') {
+            $chunks[] = $lastChunk;
+        }
+        return implode('_', array_map('mb_strtolower', $chunks));
```
TiiFuchs, line 1322: The whole method is pretty verbose for something that could literally be something like 
```
mb_strtolower(preg_replace('/(.)(?=[\p{Lu}])/u', '$1_', $value));
```

But we could rewrite that after merging.

TiiFuchs (changes requested):
I added some comments with a few thoughts.
If @noplanman agrees, those should be changed before merging.

In regards of the BC-breaking change, I think that should be ok. 
We could eventually write something into the log if there is a case, that has changed bevavior, so the user has a clue, if something is wrong all of the sudden.
I think that would be pretty smart. Log output should be removed in a later version though.

src/Telegram.php
mlocati: > Returning an empty string if no command is found, is not good style.

I usually agree, except the cases when an empty string is not a valid value (like this case).
To check if the return value is valid we'd need to check if it's not null OR is an empty string: by avoiding null we simply have to check for empty strings.

src/Telegram.php
mlocati: Yep, much better!

src/Telegram.php
mlocati: Code updated

noplanman:
Thanks @mlocati :tada: :smiley:

PR #1371, opened 2022-11-07
Bot API 6.3

TiiFuchs:
Todo in #1370

PR #1374, opened 2022-11-23
Message::getType lost some value

jyxjjj:
dice
forum_topic_created
forum_topic_closed
forum_topic_reopened

| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | no
| Fixed issues | #1371

#### Summary

These are message types, i found it added to Message::subEntities but not added to Message::getType.

I cannot confirm if there are more things forgotten.

I think it should be added, but i cannot confirm too.

If should not be added, please tell why.

### DIFF
--- a/src/Entities/Message.php
+++ b/src/Entities/Message.php
@@ -253,6 +253,7 @@ class Message extends Entity
             'location',
             'venue',
             'poll',
+            'dice',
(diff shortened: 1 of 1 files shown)


TiiFuchs:
You're right! :) Thanks a lot!

Issue #1383, opened 2023-02-18
<update> object and message is wrong for CustomCommand


uspilot:
### 🐞 Bug Report

<--
❗ NEVER put your Telegram API key or any other private details here. (like passwords, user IDs, etc.)
Substitute them like <API_KEY> or <USER_ID> etc.
-->

#### Required Information

|  ?                        |  !
| ---                       | ---
| Operating system          | Name and version
| PHP Telegram Bot version  | 0.80.0
| PHP version               | 8.1.7
| MySQL version             | 5.5.52
| Update Method             |  getUpdates
| Self-signed certificate   |  no
| RAW update (if available) | `{...}`

#### Summary

CustomCommand doesn't get new Update. CustomCommand always use the first update that has been received
#### How to reproduce

Use any CustomCommand and check <update> object  

#### Expected behaviour

<update> object must be current

uspilot:
The solution I used for me is:

telegram.php -> executeCommand function

            $this->last_command_response = $command_obj->setUpdate($this->update)->preExecute();

noplanman:
Thanks for bringing this up @uspilot 

This is definitely a breaking change, as it changes the way the Update object may have been used in users' projects.

Would you be able to share more about your use-case for this?

uspilot:
I did some censor bot which should delete messages from group chats according to defined rules. 

So, the GeneralmessageCommand class is used by default to deal with messages. Once this class is created at the first time it receives an Update object. On any later use of this class the pointer to this class has been returned without updating an Update object inside this class. So in fact you have a pointer to the class with Update object which has been sent to this class at the first use of it. So, you have a command class without any information about Update object you have to work with.

TiiFuchs:
This is solved by #1384, hopefully there will be a release soon... Stay tuned.

PR #1384, opened 2023-02-18
Fix issue #1383 with no updates for CustomCommand

uspilot:
| ?            |  !
|---           | ---
| Type         | bug 
| BC Break     | yes
| Fixed issues | #1383

### DIFF
--- a/src/Telegram.php
+++ b/src/Telegram.php
@@ -682,7 +682,7 @@ class Telegram
         } else {
             //execute() method is executed after preExecute()
             //This is to prevent executing a DB query without a valid connection
-            $this->last_command_response = $command_obj->preExecute();
+            $this->last_command_response = $command_obj->setUpdate($this->update)->preExecute();
         }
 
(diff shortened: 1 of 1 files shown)


yzen-dev:
I look forward to when this PR will be approved. Now the bot in getUpdates mode is useless. Because there is no request update, then all received requests are treated as first request.

noplanman:
Hi @uspilot  and @yzen-dev

Checking this again, I'm struggling to see the problem.

Looking at the code in this PR and testing it on my local machine, the `$command_obj` already has the same Update object as `$this->update`, so the code doesn't change anything for me :confused: 

I'd greatly appreciate it if you could provide some simple command that I can test that illustrates the problem.

Thanks!

TiiFuchs:
I'm not able to test this thoroughly, but it looks fine to me too.

Let's merge and test it with all upcoming changes for more fun and adventure!

Issue #1375, opened 2022-12-07
Switch from #[AllowDynmicProperties] to __set method for data properties


jyxjjj:
#### Required Information

|  ?                        |  !
| ---                       | ---
| Operating system          | Fedora 37
| PHP Telegram Bot version  | 0.80.0
| PHP version               | 8.2.0
| MySQL version             | none
| Update Method             | Webhook
| Self-signed certificate   | yes
| RAW update (if available) | 

#### Summary
Creation of dynamic property is deprecated

#### Current behaviour

```
$this->$key = $value;
```

```

[2022-12-07 10:53:35.971230]
channel=deprecations
level=WARNING
message=Creation of dynamic property Longman\TelegramBot\Entities\User::$id is deprecated in /www/wwwroot/TelegramBot/vendor/longman/telegram-bot/src/Entities/Entity.php on line 99
context=
extra=

```

#### How to reproduce

Use PHP 8.2.0

jyxjjj:
There are two ways to fix this warning, one is switch to enum that i talked about it months ago, one is to implement ArrayAccess, Laravel's Models used ArrayAccess then I did not get this warning.

TiiFuchs:
Currently the whole library works with dynamic properties. We didn't discuss how to handle this yet.

Implementing a `public function __set($name, $value);` and saving the properties in an array instead of a dynamic property would solve this problem too and might be the most elegant solution. We could cleanup those $raw properties in the same steps.

Or we could just add the `#[AllowDynamicProperties]` attribute to the class. (This is doesn't result in errors in earlier PHP versions!)

I don't see how Enums would help at all with the actual problem and dynamic properties.

Accessing every property via ArrayAccess would help too, but as I said earlier: The whole library works with dynamic properties, so this would be the most complex and bc-breaking solution of all. We should not do that, as long as there is another way.

I think the magic set function would be the best way to go for us.
@noplanman your thoughts on this?

jyxjjj:
yeah, I am in favor of using this scheme magic __set, but i think it will always be a temp way, finally we should define all the properties.

jyxjjj:
I read the wiki 
https://wiki.php.net/rfc/deprecate_dynamic_properties

First of all, sorry for my English, it seems like 
#[AllowDynamicProperties]
will be like a new keyword to allow using it and will not be deleted or mark as deprecated in the future?
Then it will be a good way to use it for old projects.

[...]

PR #1388, opened 2023-04-22
Feat: allow dynamic properties for PHP8.2

alesinicio:
Add `AllowDynamicProperties` attribute to Entities.php for PHP8.2 compatibility. Not a permanent/final solution, but good enough to allow upgrades.

#1375


noplanman:
@TiiFuchs Happy to do this for now?

TiiFuchs:
Yeah we had this coming. Need to add `#[AllowDynamicProperties]` for now...

TiiFuchs:
(Those checks are not very useful at all at the moment 😞)

TiiFuchs:
For the record I'l cross-reference this issue in which we discussed this topic before: https://github.com/php-telegram-bot/core/issues/1375

PR #1389, opened 2023-04-25
Bot API 6.4-6.7

noplanman:
| ?            |  !
|---           | ---
| Type         | feature / improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

The long awaited PR to update to the latest Bot API 6.7 :tada: 

@TiiFuchs Feedback welcome of course.

I'm wondering if the new entity classes should be organised differently. Like all the `Bot*` into a `Bot` folder and `InlineQuery*` ones into the existing `InlineQuery` one?

Also, I've introduced a public static variable `Entity::$fixThumbnailRename` which is `true` by default, which should make sure that bots using the old `thumb` field and method calls [don't suddenly break](https://github.com/php-telegram-bot/core/pull/1389/files#diff-72f4fd3cf2a3b01ecb2cf2ca503f9333c70a64a761cdf900f101787c2df426a0R191). It converts all found `thumb` fields and method names to `thumbnail`.

Also notice the [improved result objects returned](https://github.com/php-telegram-bot/core/pull/1389/files#diff-6389f5bdf7d2ad90ebced0b6217bf3f703bdcd51b68d0ffb23101faa530d0ad5) in the `ServerResponse` object.

I've not done extensive testing yet.


TiiFuchs (approved):
You have my respect and gratitude.
I hereby approve thee changes.

jacklul:
Seems like changes about `thumb_url -> thumbnail_url` were not applied to `InlineQueryResultPhoto` and `InlineQueryResultVideo` comments.

noplanman:
@jacklul Hmm, I wonder why, as the fixing happens on the Entity level when the properties are assigned.

Do you have some example code that shows the bug?

PR #1390, opened 2023-04-27
Refactor: better handling of dynamic properties for Entities

alesinicio:
Introduce usage of magic __set and __get methods to handle dynamic properties in Entities class.

#1375

### DIFF
--- a/src/Entities/Entity.php
+++ b/src/Entities/Entity.php
@@ -24,10 +24,10 @@ use Longman\TelegramBot\Entities\InputMedia\InputMedia;
  * @method array  getRawData()     Get the raw data passed to this entity
  * @method string getBotUsername() Return the bot name passed to this entity
  */
-#[\AllowDynamicProperties]
 abstract class Entity implements \JsonSerializable
 {
     public static $fixThumbnailRename = true;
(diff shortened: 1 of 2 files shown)


noplanman:
@TiiFuchs This should be ok to merge, yes?

Or was there something else you wanted to have in here?

TiiFuchs:
I'm not that fond of the property name being "parameters".
Telegram calls them "fields" in their documentation, PHP usually calls them "properties". Parameters are the one on functions. That's misleading.

I vote renaming them to "fields".

Additionally: Should we throw an Exception when accessing a non-defined field?
That's not exactly the previous behavior, previously you would at least get a warning. Now nothing.

An Exception would help, since most of those Classes get their fields set from the incoming API Update. So if you access $update->mssage it was 100% an error, that get's undetected.

Let's merge it but rename the field, and add a throw there. What do you think, @noplanman ?

noplanman:
@TiiFuchs Go for it!

TiiFuchs:
I added the discussed refactoring and changed the `jsonSerialize()` method as well.
See https://github.com/php-telegram-bot/core/pull/1397

PR #1399, opened 2023-05-07
Handle raw_data better in Entity class

noplanman:
| ?            |  !
|---           | ---
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

Better handling of the `raw_data` property, moving it to a dedicated property instead of sharing the main data array.


TiiFuchs (approved):
Code looks good. We should test it quite thoroughly though.

PR #1425, opened 2023-10-24
Fix building TelegramException in getCommandsList()

mlocati:
| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

This fix the following TypeError:

Exception::__construct():
Argument #2 ($code) must be of type int, Exception given


noplanman:
Nice catch @mlocati, thanks!

Issue #1456, opened 2024-04-28
Fix PHP Error - Call to a member function getChat() on null


AkramiPro:
#### Required Information

|  ?                        |  !
| ---                       | ---
| Operating system          | android 12
| PHP Telegram Bot version  | 0.82.0
| PHP version               | 8.2
| MySQL version             | none
| Update Method             | Webhook
| Self-signed certificate   | no

#### Summary

in some case maybe users want to edit command and when the commend edited the `$this->getMessage()` return null and must use `$this->getEditedMessage()` to fix this issue

#### How to reproduce

for example send `/whois 1234` and then edit the message to `/whois 5678`

#### Expected behaviour

it must resend the result base on new message edited

#### Current behaviour

```
Error - Call to a member function getChat() on null;
File: vendor/longman/telegram-bot/src/Commands/AdminCommands/WhoisCommand.php; Line: 65;
```

PR #1457, opened 2024-04-28
Fix PHP Error - Call to a member function getChat() on null

AkramiPro:
| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | yes
| Fixed issues | #1456

#### Summary

in some case maybe users want to edit command and when the commend edited the `$this->getMessage()` return null and must use `$this->getEditedMessage()` to fix this issue


noplanman:
@AkramiPro Thanks! :pray:

Issue #1446, opened 2024-01-28
Bot API 7.0


noplanman:
https://core.telegram.org/bots/api-changelog#december-29-2023

**Reactions**

- [x] Added the classes `ReactionTypeEmoji` and `ReactionTypeCustomEmoji` representing different types of reaction.
- [x] Added updates about a reaction change on a message with non-anonymous reactions, represented by the class `MessageReactionUpdated` and the field `message_reaction` in the class `Update`. The bot must explicitly allow the update to receive it.
- [x] Added updates about reaction changes on a message with anonymous reactions, represented by the class `MessageReactionCountUpdated` and the field `message_reaction_count` in the class `Update`. The bot must explicitly allow the update to receive it.
- [x] Added the method `setMessageReaction` that allows bots to react to messages.
- [x] Added the field `available_reactions` to the class `Chat`.

**Replies 2.0**

- [x] Added the ability to reply to messages in other chats or forum topics.
- [x] Added the class `ExternalReplyInfo` and the field `external_reply` of type `ExternalReplyInfo` to the class `Message`, containing information about a message that is replied to by the current message, but can be from another chat or forum topic.
- [x] Added the ability to quote a part of the replied message.
- [x] Added the class `TextQuote` and the field `quote` of type `TextQuote` to the class `Message`, which contains the part of the replied message text or caption that is quoted in the current message.
- [x] Added the class `ReplyParameters` and replaced parameters `reply_to_message_id` and `allow_sending_without_reply` in the methods `copyMessage`, `sendMessage`, `sendPhoto`, `sendVideo`, `sendAnimation`, `sendAudio`, `sendDocument`, `sendSticker`, `sendVideoNote`, `sendVoice`, `sendLocation`, `sendVenue`, `sendContact`, `sendPoll`, `sendDice`, `sendInvoice`, `sendGame`, and `sendMediaGroup` with the field `reply_parameters` of type `ReplyParameters`.

**Link Preview Customization**

- [x] Allowed to explicitly specify the URL that will be used for link preview generation in outgoing text messages.
- [x] Allowed to position link previews above the message text.
- [x] Allowed to choose media size in link previews.
- [x] Added the class `LinkPreviewOptions` and replaced the parameter `disable_web_page_preview` with `link_preview_options` in the methods `sendMessage` and `editMessageText`.
- [x] Replaced the field `disable_web_page_preview` with link_preview_options in the class `InputTextMessageContent`.
[...]

PR #1459, opened 2024-05-13
Bot API 7.0

TiiFuchs:
See changelog in #1446


Hitmare:
> I think varchar needs a size: https://mariadb.com/kb/en/varchar/

As far as i'm aware i've always had to set a size for varchar. 
In most databases VARCHAR is a fixed max size , hence why it needs a defined size

Issue #1464, opened 2024-05-24
Bot API 7.1


noplanman:
https://core.telegram.org/bots/api-changelog#february-16-2024

- [ ] Added support for the administrator rights `can_post_stories`, `can_edit_stories`, `can_delete_stories` in supergroups.
- [ ] Added the class `ChatBoostAdded` and the field `boost_added` to the class `Message` for service messages about a user boosting a chat.
- [ ] Added the field `sender_boost_count` to the class `Message`.
- [ ] Added the field `reply_to_story` to the class `Message`.
- [ ] Added the fields `chat` and `id` to the class `Story`.
- [ ] Added the field `unrestrict_boost_count` to the class `Chat`.
- [ ] Added the field `custom_emoji_sticker_set_name` to the class `Chat`.

PR #1465, opened 2024-05-24
Bot API 7.1

noplanman:
| ?            |  !
|---           | ---
| Type         | feature
| BC Break     | no
| Fixed issues | #1464 

#### Summary

Bot API 7.1 implementation


src/Entities/Message.php
```diff
@@ -38,9 +38,11 @@
  * @link https://core.telegram.org/bots/api#message
  *
  * @method int                                    getMessageId()                              Unique message identifier
- * @method int                                    getMessageThreadId()                        Optional. Unique identifier of a message thread to which the message belongs; for supergroups only
+ * @method int                                    getMessageThreadId()                        Optional. Unique identifier of a message thread to which the message belongs;
+for supergroups only
```
TiiFuchs, line 42: That newline might be a mistake...?

TiiFuchs (approved):
There might be newline that needs to be removed. 
But everything else is fine! 👍🏻
Thank you!

src/Entities/Message.php
noplanman: Ah yes, the forced newline when auto-formatting, ned to find a way to disable that. We should probably look into pint anyway for formatting.

PR #1472, opened 2024-06-25
PHPDoc comment in MessageOriginChat mislead to wrong getter getChat().

r3bers:
Factory creates getter getSender_chat() because property name sender_chat in MessageOriginChat

| ?            |  !
|---           | ---
| Type         | bug
| BC Break     | yes / no
| Fixed issues |  New. 

#### Summary
 
Just simple quick fix because PHPDoc Comment was the same to MessageOriginChat and MessageOriginChannel but properties isn't

### DIFF
--- a/src/Commands/AdminCommands/CleanupCommand.php
+++ b/src/Commands/AdminCommands/CleanupCommand.php
@@ -69,7 +69,7 @@ class CleanupCommand extends AdminCommand
      *
      * @var array
      */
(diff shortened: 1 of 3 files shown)


TiiFuchs:
It should be getSenderChat().
The snake case field names get converted to PascalCase. 

Could you adjust it accordingly, please?

r3bers:
Oh Sorry, my bad. Too much python for today. Pushed getSenderChat()

